PR #46, opened 2026-04-23
Add logEvent helper and broaden agent-id log coverage

longfin:
> Stacked on top of #45. Will rebase onto main once #45 lands.

## Summary
- Introduces `packages/server/src/log.ts` with a tiny `logEvent(event, fields)` helper that stamps `event`/`ts` and merges callers' fields. Keeps the existing console-based JSON-line format so existing log consumers (e.g. `fly logs | grep`) keep working; gives us one place to swap to a real logger later.
- Migrates the existing 6 log sites to the helper (`client_connected`, `client_disconnected`, 4× `client_rejected`, `task_unreachable`).
- Adds 5 new log points, all carrying `agentId` so they're greppable by agent:
  - `agent_request` — entry to `POST /agents/:id` (after `agentAuthMiddleware` passes)
  - `agent_request_rejected` — each reject path in `agent-auth.ts` (`reason`: `agent_not_connected` | `missing_bearer` | `bad_token_prefix` | `invalid_token` | `caller_not_authorized`)
  - `task_completed` / `task_failed_by_client` — when the connected ws client publishes a terminal frame

## Why
Today, `fly logs -a vicoop-bridge-server | grep <agentId>` only surfaces ws connection lifecycle events. There's no trace of an A2A HTTP request actually arriving, no trace of why agent-auth rejected one, and no trace of how the worker side reported the task back. That made the original \"task created with status: failed\" investigation harder than it needed to be.

## Test plan
- [ ] Send a message to a connected agent → expect `agent_request` then `task_completed` (or `task_failed_by_client`) with the same `agentId`.
- [ ] Send a message to an unknown agent id → expect `agent_request_rejected reason=agent_not_connected`.
- [ ] Send with no/invalid `Authorization` header → expect `agent_request_rejected` with `missing_bearer` / `bad_token_prefix` / `invalid_token` as appropriate.
- [ ] Send with a valid caller token that isn't in `allowed_callers` → expect `agent_request_rejected reason=caller_not_authorized` with `principalId`.
- [ ] Confirm existing `client_connected` / `client_disconnected` / `client_rejected` lines still appear with the same shape.

### DIFF
--- a/packages/server/src/agent-auth.ts
+++ b/packages/server/src/agent-auth.ts
@@ -3,6 +3,7 @@ import type { ClientConnection, Registry } from './registry.js';
 import type { Sql } from './db.js';
 import { CALLER_TOKEN_PREFIX, verifyCallerToken } from './auth/caller-token.js';
 import { matchPrincipal, type VerifiedCaller } from './auth/principal.js';
+import { logEvent } from './log.js';
 
 export function getAgentConn(c: Context): ClientConnection {
   return c.get('agentConn') as ClientConnection;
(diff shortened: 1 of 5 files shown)


longfin:
Superseded by #47 (consolidated into a single PR).

PR #56, opened 2026-04-24
Stream openclaw assistant messages as independent A2A artifacts

longfin:
## Summary

- Flips `capabilities.streaming` to `true` on the openclaw card and adds message-boundary streaming to the adapter — without any upstream OpenClaw change.
- Each assistant message appended to the transcript surfaces as its own `TaskArtifactUpdateEvent` SSE frame to the A2A caller, followed by the terminal `TaskStatusUpdateEvent(final:true)`.

## Why this route (option b: run-level multi-artifact)

True per-token deltas remain blocked on the upstream fix described in #2: OpenClaw's `chat.send` handler doesn't register a `clientRunId` in `chatRunState.registry`, so `emitChatDelta()` (which publishes via `nodeSendToSession`) never reaches an `role:\"operator\"` connection.

OpenClaw does, however, emit a `session.message` event on every transcript write to connections that called `sessions.messages.subscribe` for the sessionKey (scope `operator.read`, which we already hold). That gives us one event per completed assistant message — thinking / tool-use / final response typically arrive as separate events. Not token-level, but more than enough to qualify as A2A streaming and unblock clients today.

Mapping choice: each `session.message` becomes a new `task.artifact` with a fresh `artifactId` and `lastChunk:true`. That matches the natural grain of the data (each message is complete on arrival, not a chunk of a larger artifact) and avoids introducing an `append` field to the wire protocol.

## Changes

- `packages/client/cards/openclaw.json` — `streaming: true`.
- `packages/client/src/backends/openclaw.ts`:
  - New `sessions.messages.subscribe` call (idempotent per connection, per sessionKey). Failure is logged and non-fatal; task falls back to the single-final-artifact path.
  - Routes `session.message` events via a per-sessionKey owner map to the right in-flight task's handler.
  - Handler emits `task.artifact` per assistant message (skips `user`/`tool` entries, dedups by `messageId`, drops late events after the terminal frame).
  - Terminal `chat.final` path skips the final-result artifact when streaming already delivered content; `task.complete.status.message` still carries the final text per A2A convention.
- `packages/client/src/backends/openclaw.test.ts`:
  - Fake gateway helper auto-acks `sessions.messages.subscribe` so unrelated tests need no changes.
  - Adds 5 streaming tests: multi-message emits as distinct artifacts, no-event fallback, role filtering, `messageId` dedup, subscribe-failure graceful degradation.

All 45 tests pass (40 pre-existing + 5 new).

## Constraints / tradeoffs

- **Not token-level.** Each event is a complete transcript entry. A simple \"say hi\" run emits 1 artifact; a tool-using run emits 2–3.
- **Concurrent same-sessionKey runs aren't both streamed.** `session.message` carries no `runId` so we can't route to two tasks at once. First task wins ownership; a concurrent second task on the same contextId uses the one-shot fallback. Normal sequential A2A usage is unaffected.
- **Subscribe failure ⇒ silent fallback.** If the gateway rejects the RPC (older OpenClaw, scope denied, etc.), we log and behave like before.

The ~20-line upstream PR described in #2 remains the path to true per-token streaming; this PR unblocks the streaming capability today with zero OpenClaw changes.

## Test plan

- [x] `pnpm --filter @vicoop-bridge/client test` — 45/45 pass.
- [x] `pnpm -r build` — typecheck/build clean.
- [ ] **Manual E2E one-shot verification** using the pattern in `docs/openclaw-e2e.md`: spin up `ghcr.io/openclaw/openclaw:latest` with `CLAUDE_AI_SESSION_KEY` mounted, run a prompt that triggers tool use (e.g. \"list /tmp with bash, summarize\"), and observe multiple `task.artifact` frames with distinct `artifactId`s preceding `task.complete`. Reviewer note: this must be run once before merge; automated E2E is out of scope for this PR.

Refs: #2

### DIFF
--- a/docs/openclaw-e2e.md
+++ b/docs/openclaw-e2e.md
@@ -120,6 +120,50 @@ docker run --rm \
 response to a `chat.abort` RPC — i.e. the backend's signal-abort path
 reached the gateway correctly.
 
+## Streaming verification
+
+A second E2E exercise lives at `packages/client/scripts/e2e-openclaw-streaming.mjs`.
+It sends a tool-use-prone prompt, collects the emitted A2A frames, and
+asserts:
+
+- at least one `task.artifact` arrives before `task.complete`,
+- all `artifactId`s are distinct,
+- the terminal frame is `task.complete` with `state: "completed"`,
+- if two or more artifacts were emitted, the first precedes the
+  terminal frame in time (otherwise the assertion is skipped — a
+  single-artifact run is the documented graceful-degradation shape and
+  still satisfies the streaming contract).
+
+Run the same way as the cancel example, with the agent's auth
+configured so a real model can respond:
+
+```bash
+# Either mount a prepared auth-profiles.json into the gateway:
+docker run --rm -d --name openclaw-e2e \
+  -v "$HOME/.openclaw-e2e/auth-profiles.json:/home/node/.openclaw/agents/main/agent/auth-profiles.json:ro" \
+  ghcr.io/openclaw/openclaw:latest
+
+# ...or interactively register a provider before running the sidecar:
+docker exec -it openclaw-e2e openclaw agents add main
+
+TOKEN=$(docker exec openclaw-e2e sh -c 'cat /home/node/.openclaw/openclaw.json' \
+  | grep -oE '"token":\s*"[a-f0-9]+"' | head -1 | sed 's/.*"\([a-f0-9]*\)"/\1/')
+
+docker run --rm \
+  --network container:openclaw-e2e \
+  -v "$PWD":/w -w /w/packages/client \
+  -e OPENCLAW_GATEWAY_TOKEN="$TOKEN" \
+  node:20 \
+  node ./scripts/e2e-openclaw-streaming.mjs
+```
+
+Without auth configured the harness still passes (the gateway emits a
+single assistant "agent failed before reply" transcript entry, which
+proves the `session.message → task.artifact` wiring end-to-end), but
+you cannot verify the multi-artifact cadence a real tool-use run
+produces. Use `DEBUG=1` in the sidecar env to see every chat /
+session.message frame as it arrives.
+
 ## Teardown
 
 ```bash
--- a/packages/client/cards/openclaw.json
+++ b/packages/client/cards/openclaw.json
@@ -3,7 +3,7 @@
   "description": "General-purpose AI assistant. Can answer questions, browse the web, run shell commands, manage files, and operate tools its operator has granted. Accepts text and inline image attachments.",
   "version": "0.0.1",
   "protocolVersion": "0.3.0",
-  "capabilities": { "streaming": false },
+  "capabilities": { "streaming": true },
   "defaultInputModes": ["text/plain", "image/png", "image/jpeg", "image/gif", "image/webp"],
   "defaultOutputModes": ["text/plain"],
   "skills": [
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@vicoop-bridge/client",
-  "version": "0.2.0",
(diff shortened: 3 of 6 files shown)


longfin:
## E2E verification — complete ✅

Ran the harness against `ghcr.io/openclaw/openclaw:latest` with `anthropic/claude-opus-4-6` as the configured provider. Two prompt shapes exercised:

| Prompt | Artifacts | Total elapsed | Shape |
|---|---|---|---|
| \"List /tmp using bash, then summarize\" (thinking + direct answer, no actual tool use) | 1 | 9.6s | Single-message fallback — matches the documented graceful-degradation shape. |
| \"Use bash to run 3 commands in order, narrate between each\" (tool use forced) | **4** | 13.1s | Four distinct `artifactId`s at ~3s intervals, one per narration message between tool calls. |

Multi-artifact frame trace from the tool-use run:

```
+25ms    task.status(working)
+3845ms  task.artifact  id=cfa579fc  bytes=57
+7007ms  task.artifact  id=37e85406  bytes=97
+10107ms task.artifact  id=bc235b3e  bytes=110
+13065ms task.artifact  id=4f9137a9  bytes=156
+13066ms task.complete(completed)
```

All 4 assertions (≥1 artifact, distinct `artifactId`s, terminal `completed`, first artifact before terminal) pass.

Last checkbox on the PR test plan: ticked.

---

## Unexpected finding — possibly worth a follow-up issue

The debug trace of the E2E run showed this:

```
[openclaw] chat event: {\"runId\":\"...\",\"seq\":5,\"state\":\"delta\",\"message\":{\"content\":[{\"type\":\"text\",\"text\":\"The\"}]}}
[openclaw] chat event: {\"runId\":\"...\",\"seq\":6,\"state\":\"delta\",\"message\":{\"content\":[{\"type\":\"text\",\"text\":\"The \`/tmp\` directory contains four...\"}]}}
[openclaw] chat event: {\"runId\":\"...\",\"seq\":7,\"state\":\"delta\",...}
...
```

i.e. `chat` events with `state: \"delta\"` and monotonically-accumulating text **are arriving on our `role:\"operator\"` connection today**. Issue #2's premise — that `chat.send`'s path never registers `clientRunId` in `chatRunState.registry`, so `emitChatDelta()` never fires for us — appears to be stale against current `openclaw:latest`. Something upstream (in openclaw) seems to have wired the registration already.

Our adapter's chat finalizer currently only reacts to `final|error|aborted` states, so delta events reach us but are silently dropped (`if (evt.state === 'final' || ...)`).

### Implication

If these delta events are reliably produced by current OpenClaw, we could deliver **true token-level streaming** as a follow-up without the upstream PR #2 originally called for:

- Add a delta branch to the `taskFinalizers` handler that emits `task.artifact` with accumulating text.
- Mapping choice to settle: either same `artifactId` + `append:true` (requires adding `append` to our wire protocol + relay forwarding), or per-delta distinct artifacts (heavier on the wire but no protocol change).
- Interaction with the message-boundary path in this PR: one or the other should run, not both, to avoid double-delivery.

### Before chasing this

Worth re-validating issue #2's claim against the current OpenClaw code — specifically whether `chat.ts`'s `chat.send` handler now calls `addChatRun(chatRunState.registry, ...)` before `dispatchInboundMessage`, or whether deltas reach us via some other path (e.g., a broadcast that doesn't gate on the registry at all). The debug trace above proves they arrive; understanding *why* tells us whether it's a stable contract or an accident.

I'd suggest a new issue \"re-survey OpenClaw delta emission on `chat.send` path\" rather than reopening #2, since the fix the issue asked for may already exist.

---

## Housekeeping

- The API key used for this E2E was shared by the requester in-chat and has been **rotated** (strongly recommended; rotation is the requester's responsibility).
- Container + mounted state volume (`~/.openclaw-e2e`) were torn down post-run.

Issue #79, opened 2026-04-29
Client onboarding: replace SIWE bootstrap with OAuth device flow


longfin:
## Background

Today, bringing up a new `vicoop-client` requires the operator to hold an Ethereum EOA, install `a2a-wallet`, and sign a SIWE message to mint a caller_token, then call `registerClient` GraphQL to receive a `CLIENT_TOKEN` (`docs/install-client.md` Step 2–3). The WS hello uses that opaque `CLIENT_TOKEN`; SIWE is only the bootstrap path.

Server already implements OAuth 2.0 Device Authorization Grant with Google as the upstream IdP (`packages/server/src/auth/device-flow.ts`, `auth/google-oauth.ts`, `auth/device-ui.ts`), today issuing `vbc_caller_*` tokens persisted to the `callers` table. So the bootstrap path can switch from "sign SIWE locally" to "open a verification URL, log in with Google, poll for token" — but we should fix a pre-existing semantic wart while we're at it: today both SIWE and device flow funnel a client-data-plane operator through `callers` even though the operator is not an A2A caller. The proposal below makes client registration a first-class outcome of the device flow rather than something the operator obtains by transiently impersonating a caller.

We have no production users yet, so this is a clean replacement rather than a migration.

## Goal

Make Google device flow the primary client-onboarding path. Operators run a single `vicoop-client login` command, complete browser auth, and receive a registered client without ever holding a wallet and without minting a caller token they don't actually need.

## Design: intent-dispatched device flow

Add an `intent` to the device flow so the same approval infrastructure (device_sessions, Google OAuth, approval UI) can deliver different artifacts depending on what the operator is trying to do.

- `device_sessions.intent TEXT NOT NULL DEFAULT 'caller'` with values `'caller'` and `'client_register'`
- `POST /oauth/device/code` accepts:
  - `intent=caller` (default) — unchanged, mints a `vbc_caller_*` token
  - `intent=client_register` with `client_name` and `allowed_agent_ids` (CSV or repeated form fields) — captured into `device_sessions` alongside the device_code
- approval UI (`auth/device-ui.ts`) renders intent-specific copy:
  - caller: "Authorize \<verification_uri_complete> to call agents on your behalf"
  - client_register: "Register a new bridge client `<client_name>` with allowed agent ids `<...>` owned by your Google account"
- `POST /oauth/token` (grant_type=device_code) branches on the session's intent:
[...]

PR #82, opened 2026-04-30
Admin agent: drop eth-only gate + Google OAuth in admin UI

longfin:
PR C of #79 — stacks on PR #80. Lifts the eth-only requirement on the admin agent at `POST /` so Google-onboarded operators reach the same chat UI as wallet operators (RLS-scoped to their own rows). Admin scope itself stays wallet-only, by design.

Resolves the open thread from the discussion on #80: "OAuth login for admin-ui" was blocked by the admin agent rejecting non-eth callers; this lifts that gate and adds the corresponding Google sign-in UI.

## Summary

**Server** (`packages/server/src/http.tsx`, `admin.ts`)
- Drop the 403 at root `POST /` for non-`eth:` principals. Any verified caller token now reaches the admin agent.
- `is_admin()` (in `admin.ts` and `schema.sql`) is unchanged — still wallet-only. `list_caller_tokens` / `revoke_caller_token` and the admin-only branch of `register_client(ownerPrincipal=…)` continue to require `eth:`+`ADMIN_WALLET_ADDRESSES`.
- `buildSystemPrompt` now branches by principal kind for the "logged in as …" copy ("wallet `0x…`" vs "Google account `email@…`"). Admin scope copy unchanged.
- `buildCustomTools` takes `principalId` directly (no more `eth:` concat from a bare wallet).
- Message metadata: `_walletAddress` dropped, `_principalId` + optional `_email` passed through.

**Admin UI** (`packages/admin-ui`)
- New `GoogleAuth` component drives the bridge's `intent=caller` device flow inside a popup. The popup is opened synchronously on the click handler (popup-blocker safe), then redirected to `verification_uri_complete` once `/oauth/device/code` lands. Polling runs in the parent and stores the resulting `vbc_caller_*` to the same `auth-token` localStorage slot SIWE uses.
- New `lib/device-flow.ts` driver — fetch + poll with `slow_down` / `expired` / `cancel` semantics.
- `App.tsx` shows both `WalletAuth` and `GoogleAuth` side-by-side in signed-out state; `Sign out` button surfaces explicitly for Google-only sessions.

**Docs**
- `remote-testing.md` §0 and §5 updated to describe the new wallet-only-admin / principal-typed-chat split.

## What's still wallet-only

`is_admin()` semantics are intentionally untouched (per the original "admin stays the same" decision in #79):
- `ADMIN_WALLET_ADDRESSES` remains a CSV of bare 0x addresses
- non-`eth:` principals never match — Google operators can run their own clients but can't see other principals' clients or revoke caller tokens

If a future PR wants to make admin scope itself principal-typed (e.g. allow `google:domain:planetariumhq.com` admins), that's a deliberate follow-up separate from this PR's scope.

## Test plan

- [x] `pnpm -r typecheck` — green (4 packages)
- [x] `node --test` — 83 pass / 0 fail / 20 skipped (DB-gated, behavior unchanged)
- [ ] Live e2e: SIWE-onboarded admin chat round-trip (regression check from PR #80)
- [ ] Live e2e: Google-onboarded user chat round-trip — sign in, list own agents, attempt admin-only tool to confirm 'admin-only' guard still rejects
- [ ] CI passes

### DIFF
--- a/docs/remote-testing.md
+++ b/docs/remote-testing.md
@@ -46,8 +46,9 @@ works for non-admin callers:
   (explicit `ownerPrincipal` is admin-only, but the default is what we want).
 - `agent_policies_update` RLS is `owner OR is_admin`, and the auto-created
   policy's owner is your principal.
-- The admin agent at `POST /` gates only on "caller token has `eth:*`
-  principal", not on admin membership.
+- The admin agent at `POST /` accepts any verified caller token (`eth:*` or
+  `google:sub:*`); admin membership only widens RLS scope to all rows.
+  Non-admin callers see/manage their own clients only.
 
 ## Env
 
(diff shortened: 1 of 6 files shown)


longfin:
Combined into #83.

PR #173, opened 2026-05-14
feat(server): canonical agent card for codex-app-server backend

longfin:
Companion to #172. Adds the `codex-app-server` entry to the canonical card map (`card-resolver.ts`) and a `cards/codex-app-server.json` mirroring the existing `codex` card's capability surface (input/output modes, openai-compat URI, traceability URI).

## Why

Clients that opt into the new backend (#172) **without** supplying an inline `--card` override would otherwise hit `4013 unknown backend kind` at hello. Operators using inline cards today are unaffected.

## Test plan

- [x] `pnpm --filter @vicoop-bridge/server typecheck` clean
- [x] `pnpm --filter @vicoop-bridge/server test` — 181/181 pass (179 prior + 2 new)
- New tests: canonical card resolution for `codex-app-server`, and pinning of the openai-compat URI on it (consistent with the existing pin loop for claude/codex/openclaw).

## Notes

- Independent of #172 — can merge in either order.
- Inline cards keep priority; the inline path doesn't reach this map.

### DIFF
--- a/packages/server/src/card-resolver.test.ts
+++ b/packages/server/src/card-resolver.test.ts
@@ -35,6 +35,24 @@ test('resolves canonical codex server card from backendKind', () => {
   assert.deepEqual(result.ok && result.agentCard.defaultOutputModes, ['text/plain']);
 });
 
+test('resolves canonical codex-app-server card from backendKind', () => {
+  // The new backend shares the codex capability surface (one Codex agent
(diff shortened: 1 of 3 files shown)


longfin:
Folded into #172 — server canonical card change shipped together with the client backend for atomicity.

PR #208, opened 2026-05-18
fix(client): steer codex away from host-side scaffolding under openai-compat + recover prose-prefixed envelopes (#207)

longfin:
## Summary

Three seams against the silent text-only completion failure observed in #207, each independently verified against a deployed `codex-Mac-pr208` agent:

1. **`CODEX_LEFTOVER_TOOL_DIRECTIVE`** appended to `developerInstructions` whenever caller-side dispatch is active. Names `update_plan` and `request_user_input` — both unconditional `executors.push(...)` entries in codex-rs `spec_plan.rs` with no feature gate, so config-level disable is not possible. Tells the model they are host scaffolding outside this task's tool surface; the only legitimate outputs are a `tool_calls` envelope or a natural-language reply.

2. **`tryParseToolCallsEnvelope` relaxation** — keep the fast (clean envelope at start) path unchanged, add a slow fallback that locates a `{"tool_calls":` marker anywhere in the trimmed text, balance-matches the surrounding JSON object (respecting string literals + escapes), and parses the slice. Recovers envelopes the model prefixed with prose narration or wrapped in a code fence.

3. **Force `sandbox=workspace-write`** on `thread/start` / `thread/resume` under caller-side dispatch. The operator's configured sandbox flows into codex's built-in permissions text (developer message content[0]), and the model conflates "codex sandbox" with "caller's tool surface" — refusing the caller's writable tools when it reads "sandbox_mode is read-only". `environments: []` already removes every codex write/exec handler, so the override is cosmetic for execution but decisive for the model's reading.

Earlier commits in this branch attempted to also empty `config.mcp_servers` on `thread/start`; live test proved that path is a no-op (`mcp_connection_manager` is session/process-level — see commit `cc5c6bc`'s message). Those commits revert cleanly inside this branch.

## Live test rounds

Against `https://vicoop-bridge-server.fly.dev/agents/codex-Mac-pr208` (PR branch source via `tsx`):

| round | build state | runs | success | observed failure mode | fix landed |
|---|---|---|---|---|---|
| 1 | directive only | 4 | 4/4 | (baseline regression check) | — |
| 2 | directive | 5 | 3/5 | (026) step-2 narration without follow-through; (028) prose-prefixed envelope discarded | parser relaxation (commit `72e1cef`) |
| 3 | directive + parser | 5 | 3/5 | (030, 031) model refused on reading "sandbox_mode is `read-only`"; 0 prose-prefixed failures | sandbox override (commit `b12af01`) |
| 4 | directive + parser + sandbox | — | — | pending — daemon back up on the new build | — |

`request_user_input` stayed at 0 calls across every round (directive holding). #207 baseline emitted 0 envelopes and produced silent text-only completion; this branch produces substantive output across all successful runs.

## What this does NOT change

- `environments: []`, `config.features.*=false`, cwd handling — all preserved.
- `list_mcp_resources` / `list_mcp_resource_templates` / `read_mcp_resource` registration — gated in `spec_plan.rs` by a session/process-level `mcp_connection_manager.has_servers()` that thread-level config can't reach. Tracked as follow-up.
- `update_plan` registration — unconditional, no feature gate; directive reduces but does not eliminate.
- Non-openai-compat callers — directive + sandbox override are gated on `callerToolDispatchActive`; parser relaxation is a no-op for non-envelope text.

## Test plan

- [x] 3 new node:test cases for the sandbox override:
  - `thread/start` sends `sandbox=workspace-write` under caller-side dispatch
  - `thread/start` passes the operator sandbox through when no caller tools
  - `thread/resume` also overrides (sandbox is re-sent per turn)
- [x] 2 new node:test cases for the directive (include / omit by gate)
- [x] 5 new node:test cases for the relaxed parser (prose prefix, code fence, balanced strings, unbalanced, trailing prose)
- [x] Existing parser test updated: `'Sure! {"tool_calls":[…]}'` now returns the envelope (intentional)
- [x] Full client suite (418 tests) passes locally
- [x] `pnpm --filter @vicoop-bridge/client typecheck` clean
- [x] 3 live test rounds against `codex-Mac-pr208` — table above
- [ ] Round 4 (directive + parser + sandbox) — pending operator reproduction

## Commits

- `d0af847` initial fix (directive + ineffective `mcp_servers: {}`)
- `645d41a` refactor: broaden mcp_servers gating
- `cc5c6bc` drop ineffective mcp_servers override after live verification
- `72e1cef` relax envelope parser to recover prose-prefixed JSON
- `b12af01` force sandbox=workspace-write under caller-side dispatch

### DIFF
--- a/.changeset/codex-openai-compat-suppress-leftover-tools.md
+++ b/.changeset/codex-openai-compat-suppress-leftover-tools.md
@@ -38,6 +38,23 @@ surrounding JSON object (respecting strings + escapes), and parse the
 slice. False positives are bounded by the marker specificity + valid-JSON
 + `tool_calls`-array gate.
 
+Also overrides `sandbox` to `workspace-write` on `thread/start` /
+`thread/resume` whenever caller-side dispatch is active. A second 5-run
+re-test caught two runs (030, 031) where the model read codex's
+built-in permissions text ("`sandbox_mode` is `read-only`") and
+explicitly refused the caller's writable tools:
+
+> "this session is currently in a read-only filesystem sandbox, so I
+>  can't create `index.html`, `styles.css`, or `script.js` from here."
+
+The model never executes locally under caller-side dispatch
+(`environments: []` removes every shell / write / apply_patch handler),
+so the operator-configured sandbox value only matters for the
+permissions-text rendering. Forcing it to `workspace-write` aligns the
+text the model reads with its actual contract (emit envelopes the caller
+will execute against the caller's workspace). Non-openai-compat callers
+still get the operator-configured sandbox unchanged.
+
 Not addressed by this PR (tracked separately):
 
 - `list_mcp_resources` / `list_mcp_resource_templates` /
--- a/.changeset/codex-openai-compat-suppress-leftover-tools.md
+++ b/.changeset/codex-openai-compat-suppress-leftover-tools.md
@@ -8,14 +8,17 @@ chaining host-side scaffolding (issue #207).
 
 Two seams:
 
-- **`config.mcp_servers: {}` on `thread/start` / `thread/resume`** — empties
-  the MCP server map at thread scope, which trips the
-  `if params.mcp_tools.is_some()` gate in codex-rs `spec_plan.rs` and
-  prevents `list_mcp_resources`, `list_mcp_resource_templates`, and
+- **`config.mcp_servers: {}` on `thread/start` / `thread/resume` for any
+  openai-compat call** — empties the MCP server map at thread scope, which
+  trips the `if params.mcp_tools.is_some()` gate in codex-rs `spec_plan.rs`
+  and prevents `list_mcp_resources`, `list_mcp_resource_templates`, and
   `read_mcp_resource` from being pushed into the tool registry. In the
   observed #207 session the model burned three turns on
   `list_mcp_resources({server:"local"})` before bailing with text-only
-  output.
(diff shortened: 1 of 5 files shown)


longfin:
Closing in favour of #212 (merged), which replaces the codex backend's JSON-text envelope path with native `thread/start.dynamicTools` dispatch and explicitly states it eliminates the failure modes catalogued in #207/#208 (prose-prefixed envelopes, malformed JSON on batched calls, step-2 narration without follow-through).

Status of this PR's three seams after #212:

- **`CODEX_LEFTOVER_TOOL_DIRECTIVE`** — no longer needed. With native dispatch, the model has a real native tool surface (caller's `dynamicTools`), so the `update_plan` / `request_user_input` distraction risk that the directive countered is structurally weaker.
- **`tryParseToolCallsEnvelope` relaxation** — dead code on the codex backend (parser dropped in #212). Still relevant in principle for `claude` / `openclaw` which keep the envelope-text path, but unobserved in the field; will file a focused micro-PR if the same prose-prefixed-envelope symptom shows up there.
- **`sandbox=workspace-write` override** — #212's empirical claim of "failure modes eliminated" implies the round-3 sandbox-refusal pattern doesn't recur under native dispatch. Trusting that.

The branch carries other artefacts (test agent registration `codex-Mac-pr208`, local `~/.vicoop-pr208/` daemon config) that are part of the review-time setup, not part of the PR. Will clean those up locally.

No code from this PR worth salvaging into main as-is.

Issue #216, opened 2026-05-19
caller-tools-mcp: switch from per-task to long-lived listener


longfin:
## Context

PR #215 (#213) wires the openai-compat caller-tools dispatch path through an in-process MCP server (\`caller-tools-mcp.ts\`). The current implementation spawns **one fresh HTTP listener per A2A task**: when a task with caller tools arrives, the daemon creates a new \`createHttpServer()\` listener on a random loopback port, spins up a fresh \`@modelcontextprotocol/sdk\` \`Server\` instance, hands the URL to the spawned \`claude\` via \`--mcp-config\`, and tears the listener down in the task's \`finally\`.

This worked for the feature-correctness pass, but the lifecycle has resource-management properties that don't scale well:

- One HTTP listener + one MCP \`Server\` instance live in the daemon heap per concurrent caller-tool task.
- Random-port turnover at high concurrency produces TIME_WAIT socket accumulation and file-descriptor pressure inside the daemon process.
- Live load testing under PR #215 (Round 1, ~9 concurrent native tasks) reproduced 4 \`claude_exit_nonzero\` failures at a near-identical ~3.2s mark with empty stderr — consistent with downstream resource contention but unconfirmed.

## Goal

Replace the per-task listener model with a **single per-backend long-lived MCP server** that multiplexes all concurrent caller-tool tasks. Each claude spawn opens its own MCP session against the shared listener; the server routes \`tools/list\` and \`tools/call\` to the right A2A task based on the MCP session id.

### 1. Per-session tool-set isolation

A long-lived server is shared by N concurrent claude clients, each with their own \`tools/list\` view. The \`tools/list\` handler must look up the calling session and return only that task's caller-declared tool set.

\`@modelcontextprotocol/sdk\` exposes \`extra.sessionId\` on the \`RequestHandlerExtra\` parameter — this is the routing key.

\`\`\`ts
server.setRequestHandler(ListToolsRequestSchema, async (req, extra) => {
  const entry = sessions.get(extra.sessionId);
  if (!entry) return { tools: [] };
  return { tools: entry.tools };
});
\`\`\`

### 2. session ↔ A2A task registry

Maintain a \`Map<sessionId, { tools, onInvoke, taskId }>\` updated by the claude backend's \`handle()\`:

[...]

PR #237, opened 2026-05-20
chore(client/claude): namespace internal MCP registration keys under `_vb-`

longfin:
## Summary

- Move both bridge-internal MCP server keys under a `_vb-` ("vicoop-bridge") prefix so an operator-supplied MCP named the same generic word doesn't last-wins clobber the bridge's entry under `--mcp-config`.
- `vicoop-bridge` → `_vb-send-file`
- `caller-tools` → `_vb-caller-tools`
- `--allowedTools` (from #236) is generated from the same `mcpServers` map, so the allowlist tracks the rename automatically — model-visible tool ids become `mcp___vb-send-file__send_file` / `mcp___vb-caller-tools__<tool>` and the matching allow rules update with no extra wiring.
- Internal `McpProtocolServer({ name })` identity strings are intentionally left as-is — they don't participate in claude's namespacing and changing them could break external observers matching on those names.

Merging the two servers into one `_vb` namespace remains a follow-up tied to #216 (long-lived listener refactor for caller-tools-mcp); filing a separate issue to track that.

## Base

Branched off `fix/claude-mcp-allowedtools-235` (#236). Merge order matters — this PR's tests assume the `--allowedTools` argv from that PR. If #236 lands first this rebases cleanly onto `main`; if reordered, switch base to `main`.

## Test plan

- [x] `pnpm --filter @vicoop-bridge/client typecheck` — clean
- [x] `pnpm --filter @vicoop-bridge/client test` — 462/462 pass
- [x] All assertions on the prior `vicoop-bridge` / `caller-tools` keys updated to the new namespaced forms
- [x] `--allowedTools` regression tests (`#235`) assert the new `mcp___vb-*` tokens

### DIFF
--- a/.changeset/claude-namespace-internal-mcp.md
+++ b/.changeset/claude-namespace-internal-mcp.md
@@ -1,0 +1,26 @@
+---
+'@vicoop-bridge/client': patch
+---
+
+claude backend: namespace internal MCP server registration keys under a
+`_vb-` ("vicoop-bridge") prefix so they cannot collide with
+operator-supplied servers under the same `--mcp-config` map.
+
+The previous keys were generic enough to clash with operator names —
+`caller-tools` in particular looks like something an operator might
+name their own MCP server, and claude's `--mcp-config` JSON resolves
+collisions last-wins, silently overwriting the bridge's entry.
+
+Renames:
+
+- `vicoop-bridge` → `_vb-send-file`
+- `caller-tools` → `_vb-caller-tools`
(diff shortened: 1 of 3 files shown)


longfin:
Folded into #236 via fast-forward — both commits land together when #236 merges. GitHub auto-marked this PR as merged because its head now points at the same SHA #236's head does. Tracking the merge-the-two-servers follow-up in #238.

Issue #241, opened 2026-05-20
Architecture: openai-compat continuation turns should diff against backend session state, not reconstruct from scratch


longfin:
## The pattern we're stuck in

The openai-compat extension is **stateless by contract**: the gateway re-emits `system`, `tools`, `tool_choice`, and the full `tool_call_history` on every turn (`extensions/openai-compat/v1/README.md`).

The underlying agent runtimes (claude, codex) are **stateful by nature** — each maintains a session/thread with full conversation history, prior reasoning, prompt cache, and tool-result memory.

The bridge sits between them and currently bridges the gap by **reconstructing the runtime session from scratch on every continuation turn**:

- **claude.ts (`claude.ts:1116-1128`, `1149`)**: spawn a fresh `--session-id` per task, prepend the full `tool_call_history` as a JSON text block inside the user message, run.
- **codex.ts (post #240)**: fresh `thread/start` per task, replay the full history via `thread/inject_items` as native `function_call` / `function_call_output` items, drive a turn.

This pattern emerged as an emergency response to #175 (double-feeding when session reuse was on and gateway also re-sent history) — "the simplest fix is to never reuse the session." It's a workaround that's been load-bearing ever since.

## Symptoms it causes

This is a single architectural problem with many surface symptoms — addressing any one in isolation only papers over the others:

| Symptom | How it shows up | Where it's tracked |
|---|---|---|
| **Re-call loops on imperative prompts** | gpt-5 / Claude lose the prior "why I called this tool" reasoning between turns and re-emit the same call. #233 was the unbounded version; #240 reduced it to 1–2 extra calls but the residual remains. | #233, #240 |
| **Quadratic token cost** | Every turn re-sends the full history. N round-trips → O(N²) cumulative `prompt_tokens`. Around 50 tool calls already costs ~25K tokens *per continuation turn*. | This issue (originally) |
| **Prompt cache dead** | claude's user-message JSON changes every turn (no cache hit on history block). codex's fresh `thread/start` discards the OpenAI Responses-side cache that `previous_response_id` would normally keep warm. | This issue |
| **Encoding overhead per backend** | Each backend has its own format for the same history — claude wraps it in JSON text, codex translates to Responses-API items, openclaw needs a third encoding (#154). The translation is lossy in both directions. | #213, #154 |
[...]

PR #242, opened 2026-05-20
feat(protocol,server): forward callerPrincipal on task.assign (#241 prereq)

longfin:
## Summary

- Add optional `callerPrincipal` field to `TaskAssignFrame` (`packages/protocol/src/index.ts`) and populate it from the binding's verified `principalId` in the WS-forwarding executor (`packages/server/src/executor.ts`).
- Prerequisite for the diff-based openai-compat session reuse in #241: client backends will key per-conversation state on `(callerPrincipal, contextId)` plus a `(system, tools)` fingerprint so one caller can't resume another's backend session by guessing or sharing a `contextId`.
- **No behaviour change in this PR.** `@vicoop-bridge/client` doesn't read the field yet — that lands in the follow-up claude / codex PRs.

## Why a sibling field, not `message.metadata._principalId`?

`stripInternalMetadata` already drops `_`-prefixed keys before the frame leaves the bridge (`executor.ts:49-59`, `executor.test.ts:110-158`), so the client never sees `_principalId` today. Re-introducing it via the same channel would either reverse that strip (leaking server-internal context to downstream handlers that forward metadata) or require an exception. A typed sibling field on the frame is the cleanest path: it carries the same value to exactly one consumer (the client backend) without polluting application-facing metadata.

## Backwards compatibility

- **`PROTOCOL_VERSION` is unchanged.** `HelloFrame.version: z.literal('0.1')` still validates in both directions.
- **zod's default `z.object` strips unknown keys.** An older client parsing a newer server's frame silently drops `callerPrincipal` and behaves exactly as before.
- **Deployment is server-first**: older-server + newer-client isn't a supported configuration. The only remaining trigger for `callerPrincipal === undefined` at the client is public-agent traffic where the server intentionally omits the field; the follow-up PRs treat that case as anonymous (never resumable across turns).
- **Public-agent path**: the field is omitted entirely on the wire (not `present-but-undefined`) — explicit `hasOwnProperty === false` assertion locks this in (`executor.test.ts:268-277`).

## Test plan

- [x] `pnpm --filter @vicoop-bridge/protocol typecheck`
- [x] `pnpm --filter @vicoop-bridge/server typecheck`
- [x] `pnpm --filter @vicoop-bridge/client typecheck` (sanity — client doesn't consume the field yet but the published types must still resolve)
- [x] `pnpm --filter @vicoop-bridge/server test` — all 182 tests pass, including the three executor tests with new frame-level assertions:
  - `executor records principalId in binding and strips _principalId from outgoing WS frame` — also asserts `frame.callerPrincipal === 'eth:0xabc'`
  - `executor omits message.metadata entirely when the only entry was _principalId` — same plus `frame.callerPrincipal === 'eth:0xabc'` (proves the field rides as a sibling, not in metadata)
  - `executor binding has no principalId when message metadata is absent` — asserts both `frame.callerPrincipal === undefined` AND `hasOwnProperty === false` (public-agent path)

## Out of scope

- The (callerPrincipal, contextId) keyed cache + (system, tools) fingerprint + diff-based tool-result injection on the openai-compat path. Those land in the follow-up PRs for `claude.ts` and `codex.ts` per the review notes on #241.

## Related

- #241 — the redesign this is a prerequisite for. See [the review notes comment](https://github.com/planetarium/vicoop-bridge/issues/241#issuecomment-4497744271) for the broader strategy.

### DIFF
--- a/.changeset/task-assign-caller-principal-241.md
+++ b/.changeset/task-assign-caller-principal-241.md
@@ -1,0 +1,36 @@
+---
+'@vicoop-bridge/protocol': minor
+'@vicoop-bridge/server': minor
+---
+
+Forward the server-verified caller principal to the connected client as
+`TaskAssignFrame.callerPrincipal`. Prerequisite for the diff-based
+openai-compat session reuse redesign in #241: the client backends will key
+their per-conversation session cache on `(callerPrincipal, contextId)`
+plus a `(system, tools)` fingerprint so one caller cannot resume another's
+backend session by guessing or sharing a `contextId`.
+
+Scope is the bridge-internal server↔client WebSocket frame only. No change
+to the A2A wire shape, the openai-compat extension contract, or any
+externally-visible API. `_principalId` continues to be stripped from
+`message.metadata` before the frame leaves the bridge — the new field
+carries the same value as a typed sibling so the client never has to
+parse the internal-metadata convention.
+
+Backwards compatibility:
+
+- `callerPrincipal` is optional and omitted on the wire when the binding
+  has no principalId (public-agent path; no auth middleware ran).
+- `PROTOCOL_VERSION` is unchanged — `HelloFrame.version` literal stays at
+  `'0.1'`, so the hello handshake remains compatible in both directions.
+- zod's default unknown-key strip on `z.object` means an older client
+  parsing a newer server's frame silently drops the new field and behaves
+  exactly as before.
+- A newer client parsing an older server's frame sees
+  `callerPrincipal === undefined` and treats the task as anonymous (never
+  resumable across turns), which is the same fail-safe path used for the
+  public-agent flow.
+
+No behaviour change in this release — `@vicoop-bridge/client` does not yet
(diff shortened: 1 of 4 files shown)


longfin:
Closing in favour of the simpler approach surfaced while reviewing this PR. The hijack risk that motivated threading `callerPrincipal` to the client turned out to be theoretical: A2A's `contextId` is UUID v4 in practice (122-bit entropy), so blind-guess collisions are negligible and the remaining leak vectors (caller-internal storage, server log access, TLS-broken transport) aren't the bridge's responsibility. Combined with a `(system, tools)` fingerprint check for cache validity, `contextId` alone is a sufficient key for the diff-based openai-compat session cache in #241.

That means the follow-up PRs (claude, codex) need no protocol change — they consume the existing `TaskAssignFrame.contextId` directly. If `callerPrincipal` on the wire becomes useful later (audit logging, anonymous-bucket isolation, debugging correlation), the same 22-line shape can be reintroduced then.

Issue #233, opened 2026-05-20
Codex openai-compat loop keeps requesting same tool despite matched tool_call_history


longfin:
## Problem

While testing `codex-Mac` through the public bridge, the caller side appeared to enter a repeated tool-call loop for a simple request:

```text
랜덤 숫자 아무거나 jot 로 찍어봐
```

The sender was stopped, after which no new `task.assign` logs arrived. The last task stored on the bridge shows that the request had already accumulated six successful `jot` tool results in `tool_call_history`, but the Codex backend still returned another `bash` tool call instead of producing a final answer.

This does not look like a tool-call ID mismatch: every assistant tool call id matched the corresponding tool result `tool_call_id` 1:1.

## Observed daemon log

The client was running `@vicoop-bridge/client@0.19.0` as `codex-Mac`.

The tail of the daemon log showed repeated tasks for the same effective request, then the sender was stopped and no further logs arrived:

```text
[client] task.assign taskId=d25754f6-e55d-41ce-bbcf-dfdc25c5d4ac ...
[client] task.complete taskId=d25754f6-e55d-41ce-bbcf-dfdc25c5d4ac elapsedMs=4821 artifacts=1
[client] task.assign taskId=73023391-dfa5-44de-b75d-8615b8c5a855 ...
[client] task.complete taskId=73023391-dfa5-44de-b75d-8615b8c5a855 elapsedMs=7987 artifacts=2
...
[client] task.assign taskId=2e03b3d4-e714-460f-97a1-bcb5ce60847d contextId=fa65744f-51be-4ddb-ae8d-3459ce0187bc parts=text/plain
[client] backend.start taskId=2e03b3d4-e714-460f-97a1-bcb5ce60847d backend=codex
[client] task.complete taskId=2e03b3d4-e714-460f-97a1-bcb5ce60847d elapsedMs=5548 artifacts=2
```

## Last task dump summary

Queried with JSON-RPC `tasks/get`:

```json
{
  "method": "tasks/get",
  "params": {
    "id": "2e03b3d4-e714-460f-97a1-bcb5ce60847d"
  }
}
```

Task summary:

```json
{
  "id": "2e03b3d4-e714-460f-97a1-bcb5ce60847d",
  "contextId": "fa65744f-51be-4ddb-ae8d-3459ce0187bc",
  "status": {
    "state": "completed"
  },
  "history": [
    {
      "role": "user",
      "parts": [
        {
          "kind": "text",
[...]

Issue #213, opened 2026-05-19
Spike: native function-call surface for openai-compat caller tools on the claude backend (#209 follow-up)


longfin:
## Context

PR #212 (#209) replaces the JSON-text envelope contract on the **codex** backend with codex's native `dynamicTools` / `item/tool/call` surface. The same envelope-text path is still in use on the **claude** backend (`packages/client/src/backends/claude.ts`) and exhibits the same architectural fragility:

- prose-prefixed envelopes (`tryParseToolCallsEnvelope` requires the assistant text to start with `{`)
- malformed JSON across batched calls (missing braces in multi-call emissions)
- step-2 narration without follow-through (the model declares "writing now" and ends the turn)
- generic "model just answered in prose instead of emitting the envelope" stochastic refusals

This issue tracks a **research spike** mirroring #209, scoped to the claude backend.

## Constraint

Claude code has **no equivalent of codex's `dynamicTools`**. The only way to put caller-supplied tools in the model's native tool surface is via `--mcp-config` (an MCP server claude connects to on spawn). So the claude analog of PR #212 has to go through MCP impersonation rather than a direct protocol field.

The good news: the infrastructure for an in-process MCP server already exists in this repo — `packages/client/src/backends/send-file-mcp.ts` is a working pattern of a per-backend loopback HTTP MCP server with active-task routing. Generalizing it to expose caller-declared tools is the central design move.

## Goal

Determine whether the claude backend can expose caller-provided openai-compat `tools` as **native MCP tools** (visible to the model as `tool_use` blocks, not envelope JSON the model has to emit as text), and prototype the most viable approach.

## Proposed sketch (Phase A — matches PR #212's multi-turn flow)

Mirror PR #212 as closely as the claude surface allows:

| codex (PR #212) | claude equivalent |
|---|---|
| `thread/start.dynamicTools` | `--mcp-config` + in-process MCP server dynamically registering caller tools per task |
| `item/tool/call` server request | stream-json `tool_use` block, or MCP `tools/call` handler |
| `turn/interrupt` + `success:false` | SIGTERM the claude child after first matching `tool_use` (or MCP handler returns `isError:true` so claude ends turn naturally) |
| `interrupted` → `completed` mapping | killed-after-tool-use → `completed` mapping (`capturedToolCall` flag) |
| `thread/resume` + `thread/inject_items` | `--resume <session-id>` + existing `formatToolCallHistory` text prepend |
[...]

PR #243, opened 2026-05-20
feat(client/claude): reuse session + diff-inject new tool results on openai-compat continuation (#241)

longfin:
## Summary

Implements the diff-based openai-compat session reuse described in #241, for the claude backend.

- Drop the `openaiCompat === null && sessionTtlMs > 0` gate at `claude.ts:1149` so openai-compat continuation turns also resume via `--resume <sid>` instead of spawning fresh every time.
- Add an `ackSet: Set<string>` to each cached `SessionEntry`. Continuation turns prepend only `role:"tool"` history entries whose `tool_call_id` isn't already in the ackSet — the matching `assistant.tool_calls` are in claude's own session memory from the earlier turn, and re-feeding them is precisely the #233 / #175 failure mode.
- Update the ackSet only on a successful terminal frame (state `completed` or `input-required`), so a failed turn leaves the invariant unchanged and the retry sees the same diff.

First turns and fresh-fallback paths (no cached entry, TTL-evicted, in-process restart) render the full history exactly as before — a caller's first sight of any given conversation is unchanged.

## What's deliberately NOT in this PR

Surfaced during the review thread on #241 (see [the strategy comment](https://github.com/planetarium/vicoop-bridge/issues/241#issuecomment-4497744271) and the closed [PR #242](https://github.com/planetarium/vicoop-bridge/pull/242)). All three were on the original plan and dropped as YAGNI / unjustified complexity once we walked through the failure modes:

- **`callerPrincipal`-keyed cache.** A2A's `contextId` is UUID v4 in practice (122-bit entropy); blind-guess hijack collisions are negligible and the remaining leak vectors (caller-internal storage, TLS-broken transport, server log access) aren't the bridge's responsibility.
- **`(system, tools)` fingerprint check.** If a caller changes their system prompt or tools array mid-conversation, the cached claude session has stale dispatch surface because `--resume` can't reissue `--append-system-prompt`. Trade-off accepted: callers that change tools should also change contextId (the natural "new conversation" boundary). Add later if production measurement shows real stale-dispatch incidents — it's a 16-hex hash and a mismatch branch, easy to bolt on.
- **Fork detection (incoming history prefix vs cached `historyPrefix`).** A caller cutting their conversation history mid-flight is malformed use of the extension. The ackSet's "drop already-delivered ids" semantics is safe even under this scenario — the worst case is the model has a stale view of an earlier turn that the caller is now claiming didn't happen.

## Operational requirement

Same-`contextId` continuation turns must land on the same bridge instance for the optimisation to apply. Round-robin across multiple bridge processes correctly falls back to fresh every turn (no regression) but realises none of the savings — sticky routing on `contextId` is required to hit the acceptance-criteria benchmark in #241.

## Test plan

- [x] `pnpm --filter @vicoop-bridge/client typecheck`
- [x] `pnpm --filter @vicoop-bridge/client test` — 469/469 pass, including the three new tests + the existing `#213`-era invariant replaced inline:
  - `openai-compat continuation: second turn with same contextId resumes via --resume (#241)` — argv-level proof: `--resume <sid>` carries the first turn's minted session id, `--session-id` absent on the second spawn.
  - `openai-compat continuation: prepended history contains only new role:"tool" results since the prior turn (#241)` — three-turn scenario verifying the prepended block contains only newly-acked tool results, never the matching `assistant.tool_calls` and never an entry whose id was already acked on a prior turn.
  - `openai-compat continuation: regression guard against #175 double-feeding` — structural invariant: across both turns' stdin payloads combined, any given `tool_call_id` appears at most once.

## Follow-up

- **PR for `codex.ts`** mirrors this pattern (#241's codex deliverable). Tracking that as the next PR — it also gets to revert the #240 session-reuse opt-out (`codex.ts:815`).
- **#216** (caller-tools MCP listener lifecycle) is unaffected by this PR — MCPs remain per-task even on a resumed session, per the issue's "out of scope" call.

## Related

- #241 — the redesign this implements
- #233, #175 — earlier symptoms of the underlying invariant violation
- #240 — codex-side counterpart workaround (this PR's pattern unblocks reverting it)
- closed PR #242 — the `callerPrincipal` protocol prerequisite that turned out to be unnecessary

### DIFF
--- a/.changeset/claude-openai-compat-session-reuse-241.md
+++ b/.changeset/claude-openai-compat-session-reuse-241.md
@@ -1,0 +1,93 @@
+---
+'@vicoop-bridge/client': minor
+---
+
+feat(client/claude): reuse the claude session across openai-compat
+continuation turns and prepend only newly-arrived `role:"tool"` results
+(#241).
+
+Before this change, every A2A task carrying the openai-compat extension
+spawned a fresh `--session-id` and the bridge re-fed the entire
+`tool_call_history` JSON block on every turn. That made the per-turn
+prompt cost grow linearly in the number of prior round-trips and the
+cumulative cost grow quadratically; it also discarded the prior
+reasoning continuity that `--resume` would otherwise carry in claude's
+own session memory. The fresh-spawn gate (`sessionReuseEligible =
+openaiCompat === null && sessionTtlMs > 0` at `claude.ts:1149`) was the
+emergency workaround for #175 — when session reuse was on AND the
+gateway also re-sent the full history, the model saw every prior round
+twice and looped.
+
+The redesign keeps the contextId-keyed session alive across openai-compat
+A2A tasks and reconciles the two sources of truth instead of avoiding
+the conflict. Each `SessionEntry` now carries an `ackSet:
+Set<string>` of `tool_call_id` values whose `role:"tool"` result has
+already been delivered to this claude session. A continuation turn:
+
+1. Resumes the prior session via `--resume <sid>` (`isResume = true`).
+2. Filters the inbound `tool_call_history` to `role:"tool"` entries
+   whose `tool_call_id` isn't in the cached `ackSet`. The matching
+   `assistant.tool_calls` entries are *not* re-fed — they're already in
+   claude's own session memory from the prior turn, and re-feeding them
+   is precisely the #233 / #175 failure mode.
+3. Renders the filtered list through the existing `formatToolCallHistory`
+   wrapper and prepends it to the user content.
+4. On a successful terminal frame (`task.complete` with state
+   `completed` or `input-required`), adds every `role:"tool"`
+   `tool_call_id` from this turn's history to the ackSet so the next
+   continuation turn doesn't replay them. Failure leaves the ackSet
+   unchanged so the retry sees the same diff.
+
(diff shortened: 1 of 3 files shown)


longfin:
Closing without merge.

Spun this branch up against production and surveyed the oai2a2a gateway code (`packages/server/src/chat-completions.ts:161` calls `buildSendMessageParams(req, { agentCard })` without a `contextId` opt, and `responses.ts:124-130` outright rejects `previous_response_id` as unsupported). The diff-based session reuse this PR implements therefore has no path to fire in practice: every `/v1/chat/completions` request lands at the bridge with a freshly-minted `contextId`, so the resume branch never runs regardless of how callers (opencode etc.) batch their turns.

#241's premise ("Tasks sharing a contextId") doesn't hold under the current gateway. Three plausible upstream prerequisites — stable contextId derivation at the gateway, `previous_response_id` support + mapping, or callers opting into `req.a2a.context_id` — are all outside the bridge.

Holding this branch open banks no value (no traffic exercises the new path) and regression risk on a long-lived branch only grows. Reopen or fork from `feat/241-claude-diff-session-reuse` once a gateway-side prerequisite lands and the test suite here can re-validate against real multi-turn traffic.

PR #278, opened 2026-05-26
feat(install): auto-add INSTALL_DIR to PATH (closes #277)

### LINKED ISSUE #277: Document PATH setup after installing vicoop-client
## Background

`docs/install-client.md` installs the binary into an app-specific `INSTALL_DIR` such as `$HOME/vicoop-bridge-client`, then most examples run it as `"$INSTALL_DIR/vicoop-client"`.

That works, but it differs from common CLI install expectations: after installing a CLI named `vicoop-client`, users naturally try `vicoop-client --version` or `vicoop-client -v`. On a fresh shell this fails unless the install directory is already on `PATH`.

The current doc also mixes forms in later examples. Most commands use `"$INSTALL_DIR/vicoop-client"`, but some troubleshooting/inspection examples use bare `vicoop-client`, which makes the expected shell setup ambiguous.

## Suggested change

Add an explicit post-install PATH section, for example:

```sh
export PATH="$HOME/vicoop-bridge-client:$PATH"
```

and shell-specific persistence examples:

```sh
# zsh
echo 'export PATH="$HOME/vicoop-bridge-client:$PATH"' >> ~/.zshrc

# bash
echo 'export PATH="$HOME/vicoop-bridge-client:$PATH"' >> ~/.bashrc
```

The installer output could also print a short next-step hint after a successful install:

```text
Installed vicoop-client to $INSTALL_DIR/vicoop-client

To run vicoop-client from any shell, add this to your shell config:
  export PATH="$INSTALL_DIR:$PATH"

Verify with:
  vicoop-client -v
```

## Notes

I do not think the installer needs to automatically edit shell rc files. That can be surprising and shell-dependent. A clear doc section plus installer hint should be enough and matches common CLI installer behavior.

longfin:
## Summary

- `install.sh` now auto-edits the operator's login-shell rc file so
  bare `vicoop-client` resolves from any new terminal:
  - `zsh` → `${ZDOTDIR:-$HOME}/.zshrc`
  - `bash` (macOS) → `~/.bash_profile`
  - `bash` (Linux) → `~/.bashrc`
  - `fish` → `${XDG_CONFIG_HOME:-~/.config}/fish/config.fish` (uses `fish_add_path`)
- `NO_MODIFY_PATH=1` opts out. Unknown shells (tcsh / ksh / …) and
  non-writable rc files automatically fall back to a printed manual
  instruction — no env override needed.
- Marker comment `# vicoop-bridge-client (vicoop-client)` makes
  re-runs (including `FORCE=1`) idempotent.
- When `$INSTALL_DIR` sits under `$HOME`, the appended line uses a
  literal `$HOME/...` prefix so it stays portable across machines
  for the same operator.
- `docs/install-client.md`:
  - New Step 1b documents the per-shell rc edit + idempotency
    + opt-out. `NO_MODIFY_PATH` row added to the env table.
  - 27 user-command examples drop the `"$INSTALL_DIR/vicoop-client"`
    prefix in favor of bare `vicoop-client`. The 4 remaining
    `$INSTALL_DIR/vicoop-client` references (on-disk layout, upgrade
    atomic-swap mechanics, the opt-out hint itself) are intentional.

Approach modeled on Bun / uv / fnm install scripts — auto-edit
default + opt-out flag + automatic print-only fallback for shells
that don't fit a known rc-file convention. mise-style print-only was
the original ask in #277, but per maintainer guidance we went
further and auto-edit by default.

Closes #277.

## Test plan

- [x] `sh -n install.sh` — POSIX shell syntax clean.
- [x] Isolated test harness exercised 7 cases:
  - fresh zsh on macOS → `.zshrc` gains marker + `export PATH` line,
    `$HOME` substituted into the path.
  - re-run on same `.zshrc` → marker detected, no duplicate append
    (file line-count stays unchanged).
  - fish → `config.fish` gains marker + `fish_add_path`, parent dir
    auto-created.
  - bash on Linux → `.bashrc` (not `.bash_profile`).
  - unknown shell tcsh → no rc edit; manual instruction recorded.
  - `NO_MODIFY_PATH=1` → no rc edit, no manual hint, log line
    explains skip.
  - `INSTALL_DIR=/opt/vicoop-bridge-client` (outside `$HOME`) →
    absolute path written verbatim, no `$HOME` substitution.
- [ ] End-to-end smoke against a fresh shell on macOS + Linux —
  recommend reviewer runs the public one-liner once before merge
  to confirm the rendered Next-steps block reads naturally.

### DIFF
--- a/docs/install-client.md
+++ b/docs/install-client.md
@@ -117,10 +117,12 @@ the matching rc file so `vicoop-client` works from any new terminal:
 | `fish` | `${XDG_CONFIG_HOME:-~/.config}/fish/config.fish` | `fish_add_path $INSTALL_DIR` |
 
 Each entry is prefixed with a `# vicoop-bridge-client (vicoop-client)`
-marker comment so re-running `install.sh` (including with `FORCE=1`)
-detects the existing line and won't duplicate it. If `$INSTALL_DIR` sits
-under `$HOME`, the appended line is written with a literal `$HOME/...`
-prefix so it stays portable across machines for the same operator.
+marker comment. Re-running `install.sh` (including with `FORCE=1` or with
+a different `$INSTALL_DIR`) strips the existing marker block and
+re-appends it with the current target so the rc file never accumulates
+duplicates or stale paths. If `$INSTALL_DIR` sits under `$HOME`, the
+appended line is written with a literal `$HOME/...` prefix so it stays
+portable across machines for the same operator.
 
 To pick the change up in the shell where you just ran `install.sh`, either
 open a new terminal or `source` the rc file the installer prints in its
@@ -140,8 +142,9 @@ with `"$INSTALL_DIR/"` — e.g. `"$INSTALL_DIR/vicoop-client" --version`.
 Set `NO_MODIFY_PATH=1` on the same `install.sh` invocation if you'd rather
 manage `PATH` yourself — recommended on NixOS, immutable OSes, or when a
 dotfile manager (chezmoi, yadm, etc.) owns your shell rc files. The
-installer prints the exact line it would have appended so you can place it
-wherever your dotfile setup expects:
+installer prints both the exact line it would have appended **and** the
+shell-specific rc file it would have appended to, so you can route the
+line wherever your dotfile setup expects:
 
 ```sh
 curl -fsSL https://raw.githubusercontent.com/planetarium/vicoop-bridge/main/install.sh \
@@ -172,14 +175,17 @@ If you'd rather keep the old install around, pick a different
 new bundle there instead.
 
 Before continuing, verify that the installed bundle is recent enough to
-include the `login` command used in Step 4:
+include the `auth login` command used in Step 4:
 
 ```sh
 vicoop-client -v
-vicoop-client login --help
+vicoop-client auth login --help
 ```
 
-If `login --help` prints usage, you're on a current binary and can move on.
+If `auth login --help` prints usage, you're on a current binary and can
+move on. The flat `login` alias still works on current binaries but emits a
+`[deprecated] Use auth login` notice to stderr — see the legacy aliases
+note in Step 4.
 
 ## Step 3 — Pick an agent id
 
--- a/docs/install-client.md
+++ b/docs/install-client.md
@@ -79,6 +79,7 @@ would still default to `/data/vicoop-bridge-client`.
 | `INSTALL_DIR` | `/data/vicoop-bridge-client` | Target directory. Pick a writable path on a volume that survives restarts. |
 | `VERSION` | latest `@vicoop-bridge/client@*` | Pin a specific tag, e.g. `@vicoop-bridge/client@0.1.1`. With `VERSION` set, `jq` is no longer required. |
 | `FORCE` | `0` | If `1`, overwrite a non-empty `INSTALL_DIR`. |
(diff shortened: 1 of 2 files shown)


longfin:
Reviewed the PATH auto-edit approach. Since default rc-file editing is approved, I think the remaining issues are implementation/docs consistency rather than the overall direction:

1. `NO_MODIFY_PATH=1` docs and behavior disagree.

   `docs/install-client.md` says the installer prints the exact line it would have appended when opting out, but `install.sh` returns early for `NO_MODIFY_PATH=1` and only later prints the absolute-path prefix fallback. Either the docs should describe the current behavior, or the installer should still compute and print the manual PATH line while skipping the rc edit.

2. Marker-only idempotency can leave PATH pointing at an old `INSTALL_DIR`.

   `install.sh` skips as soon as `# vicoop-bridge-client (vicoop-client)` exists, without checking whether the following PATH entry matches the current `INSTALL_DIR`. If a user reinstalls to a different path, such as moving from `$HOME/vicoop-bridge-client` to `/opt/vicoop-bridge-client`, the rc file keeps the stale entry and bare `vicoop-client` may keep resolving to the old binary or not resolve at all. The marker block should probably be updated/replaced when the target path changes, or the marker should encode enough path/version context to avoid this stale skip.

3. Step 2 still validates the deprecated alias.

   `docs/install-client.md` now uses `auth login` in the installer next-steps, but Step 2 still says `vicoop-client login --help`. Current binaries print `[deprecated] Use auth login` there, so this should be `vicoop-client auth login --help` for consistency.

longfin:
Thanks for the careful read — all three addressed in f6b8352.

1. **`NO_MODIFY_PATH=1` parity.** Moved shell detection above the
   opt-out branch in `install.sh`, so even when we skip the rc edit
   we still populate `PATH_MANUAL_HINT` *and* `PATH_MANUAL_TARGET`
   (the shell-appropriate rc path). The Next-steps block now reads
   "add this line to `<rc-file>` (or wherever your dotfile setup
   expects): `<line>`". Unknown shells (tcsh / ksh / …) keep just the
   line without a target, since we don't know which file is right.
   The trailing `Opted out…` fallback block is now unreachable and
   was removed.

2. **Marker idempotency → strip-and-replace.** On every re-run the
   installer now finds the marker block (optional leading blank +
   marker + PATH/fish_add_path line), strips it via a one-pass `awk`
   with single-line buffering, and re-appends with the current
   `install_dir_expr`. Side effects:
   - Same `$INSTALL_DIR` re-run → no duplicate; file line count is
     stable. Log line distinguishes `added` vs `refreshed`.
   - Different `$INSTALL_DIR` re-run → old line removed, new path
     written. The "stale marker → wrong binary" hazard is gone.
   - User content around the marker is preserved (Case G in the test
     harness writes a multi-line `.zshrc` with `export FOO=bar` +
     `alias ll`, runs install twice with different `INSTALL_DIR`s,
     verifies both user lines survive and exactly one marker remains).
   - Empty-content edge case: `awk` output empty (rc had only the
     marker block) is handled by truncating to zero bytes instead of
     `printf '%s\n' ""` which would write a stray `\n` and accumulate
     blanks across re-runs.
   - In-place rewrite via `> "$rc_file"` preserves the file's
     existing mode (matters for rc files at 0600).

3. **`auth login --help` in Step 2.** Updated. Added a one-liner that
   acknowledges the flat `login` alias still works on current binaries
   but emits `[deprecated] Use auth login`, pointing readers at the
   existing legacy-aliases note in Step 4.

Doc copy in Step 1b was updated to reflect the new strip-and-replace
behavior ("re-running … strips the existing marker block and
re-appends it with the current target so the rc file never accumulates
duplicates or stale paths") and the richer opt-out message ("both the
exact line **and** the shell-specific rc file").

Issue #200, opened 2026-05-18
openai-compat envelope is duplicated as both data artifact AND text in status.message.parts


longfin:
## Summary

When the `openai-compat/v1` extension is active and the model emits a `{"tool_calls":[…]}` envelope, all three backends (`codex`, `claude`, `openclaw`) emit the **same envelope twice** in the same A2A task:

1. As a `data` part on a `task.artifact` (correct — `data: { tool_calls: […] }`).
2. As a `text` part on the terminal `task.complete.status.message` (the raw envelope JSON string).

The upstream gateway (`oai2a2a`) consumes **both** — it forwards the artifact's `tool_calls` once, then re-extracts the same `tool_calls` from the status message text via `parseToolCallsFromText` and emits them a second time over the OpenAI streaming response. With OpenAI's streaming protocol the client accumulates `tool_calls[].function.arguments` by index, so the duplicated emission produces invalid JSON like `{…}{…}` and the OpenAI client (opencode) silently drops or loops on the call.

This is the root cause of planetarium/oai2a2a#50 (write tool called 5× in a row) and planetarium/oai2a2a#51 (tool_calls not delivered to client).

## Confirmed task data

Fetched from a real failing task via `a2a-wallet a2a tasks get`:

```
task 44a9a1df-0fb2-41e1-9081-5ecccead830b
status.state: completed
artifacts: 5
  [0..3] codex-message  kind=text   (preamble messages)
  [4]    codex-message  kind=data   ext=[openai-compat/v1]
         data.tool_calls = [
           { id: "call_write_index",  function.name: "write" },
           { id: "call_write_styles", function.name: "write" },
           { id: "call_write_script", function.name: "write" }
         ]

status.message.parts = [
  { kind: "text", text: '{"tool_calls":[call_write_index, call_write_styles, call_write_script]}' }
]
```

The envelope text on `status.message.parts[0]` is byte-for-byte the same JSON whose parsed form is already carried by artifact [4]'s data part.

## Root cause in code

All three backends do the same thing: after emitting the envelope as a `data` artifact (which sets `emittedAnyArtifact = true`), they still build the terminal status message from the **raw** agent text (`finalText` / `completeText`) without checking whether that text is itself the envelope already routed via a data artifact.

### `packages/client/src/backends/codex.ts:1089-1136`

```ts
const completeText = outcome.finalText ?? '';
const parts: Part[] = completeText
[...]

PR #296, opened 2026-05-27
feat(client): stream vicoop-codex backend via long-running serve child

longfin:
## Summary

- Replace the `vicoop-codex` backend's subprocess-per-task model with a long-running `vicoop-codex serve --port 0` child supervised on the bridge side.
- New `ServeChildSupervisor` (modeled on `AppServerRpcClient`): spawn once, parse the JSON listening banner the CLI prints on its first stderr line to discover the bound port, lazy spawn / crash respawn via `waitForClose`, killed on shutdown through the `cli.ts` hook.
- Each A2A task POSTs `/v1/chat/completions` with `stream: true`; SSE chunks are translated to token-level `task.artifact` frames (`append: true`, terminated with `lastChunk: true`).
- Tool-call deltas are accumulated across chunks (name on first delta, arguments concatenated across the rest) and emitted as a single `data` artifact under the openai-compat extension URI — same wire shape as before so the oai2a2a gateway keeps decoding it.
- Terminal `task.complete` carries the same `chat_completion` echo + `usage` metadata the previous non-streaming path produced.

## Dependency

Requires the `vicoop-codex` CLI release that supports `--port 0` and emits the JSON listening banner — planetarium/vicoop-codex-cli#5. Until that ships, `vicoop-codex` ≥ the version from that PR must be installed locally (`npm link` works for dev).

## Why

Issue planetarium/vicoop-codex-cli#4 — bridge's previous `vicoop-codex call` path buffered the full Chat Completions response and emitted one final artifact, so OpenAI-compat clients couldn't receive incremental SSE chunks. After the CLI option-comparison thread (#4 comments), `serve` came out the simpler path: streaming was already implemented there, and the bridge already had a long-running-subprocess supervisor pattern (codex's `app-server`) to model on.

## Test plan

- [x] 23 tests pass (`node --import tsx --test src/backends/vicoop-codex.test.ts`) — pure helpers (`flattenA2AUserContent`, `buildMessages`, `buildCallBody`, `parseChatCompletionUsage`, `buildResponseMetadata`) plus integration tests covering text streaming, tool_calls assembly across deltas, body shape, HTTP 401 / 429 / 502 mapping, mid-stream `error` event, cancel-before-spawn, startup timeout, supervisor reuse across tasks, and crash respawn.
- [x] `pnpm -r typecheck` clean.
- [x] End-to-end against real ChatGPT (via `npm link`-installed CLI from planetarium/vicoop-codex-cli#5): a 5-token prompt streamed 10 artifact frames (first delta +2916ms, subsequent ~50ms gaps, terminal `lastChunk:true` frame, `task.complete` with `chat_completion` metadata echo).
- [ ] Multi-turn tool-loop scenario validated against a real backend (covered by unit tests; smoke-tested manually with a single-turn synthetic).

### DIFF
--- a/.changeset/vicoop-codex-http-streaming.md
+++ b/.changeset/vicoop-codex-http-streaming.md
@@ -1,0 +1,5 @@
+---
+"@vicoop-bridge/client": minor
+---
+
+Replace the `vicoop-codex` backend's subprocess-per-task model with a long-running `vicoop-codex serve --port 0` child supervised on the bridge side. Each A2A task POSTs `/v1/chat/completions` with `stream: true` and emits token-level artifact deltas, plus a terminal `chat_completion` echo on `task.complete` for OpenAI-compat callers. Requires the `vicoop-codex` CLI release that ships `--port 0` support and the JSON listening banner (planetarium/vicoop-codex-cli#5).
--- a/packages/client/src/backends/vicoop-codex-supervisor.ts
+++ b/packages/client/src/backends/vicoop-codex-supervisor.ts
@@ -1,0 +1,279 @@
+import { spawn as nodeSpawn, type ChildProcess } from 'node:child_process';
+import { createLogger, type Logger } from '../logger.js';
+
+// Slim subset of ChildProcess the supervisor uses. Tests inject a fake.
+export interface VicoopCodexChildHandle {
+  readonly stdin: NodeJS.WritableStream | null;
+  readonly stdout: NodeJS.ReadableStream | null;
+  readonly stderr: NodeJS.ReadableStream | null;
+  kill(signal?: NodeJS.Signals): boolean;
+  on(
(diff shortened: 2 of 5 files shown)


longfin:
Superseded by #315, which lands the same approach (long-running `vicoop-codex serve` singleton + token-level `append:true` SSE streaming) rebased on current main.

Key differences vs this PR:
- tool_calls now flow **only** through the terminal `chat_completion` envelope per the oai2a2a#80 envelope contract (#200/#302), not as a `data` artifact — this PR's data-part wire is stale.
- The `vicoop-codex` CLI `serve` support it depended on has since shipped, so #315 targets the released binary (`serve -p 0`, JSON listening line on stderr).
- Bundled agent card flipped to `capabilities.streaming: true`.

Closing in favor of #315.

PR #328, opened 2026-06-05
fix(client): give each caller-tool call a conversation-unique id

longfin:
## Problem

The claude backend derived the OpenAI `tool_call_id` from the MCP JSON-RPC request id (`extra.requestId`) in `caller-tools-mcp.ts`. That id is only monotonic **within a single MCP session**, but the bridge stands up a brand-new caller-tools MCP server (and claude process) **per A2A turn**, so the counter reset every turn:

- first `tools/call` of each turn → `2`
- further calls in the same turn → `3,4,5…`

When the openai-compat loop replayed the accumulated `<chat_history>`, those non-unique ids collided across turns. In one real session, `tool_call_id` "2" appeared **15 times** across 25 calls (only 5 distinct ids: 2,3,4,5,6).

## Symptom

With colliding ids, the call↔result pairing breaks (and the system-prompt guard *"don't repeat a call whose tool_call_id already appears"* is defeated). The model loses track of what it called vs. what came back, so it restates its entire plan before every tool call — the repeated *"I'll wire the visual novel… update the docs"* preambles visible in the backend prompt view around each tool call.

## Fix

Always mint a fresh `call_<uuid>`. The id only needs to be unique and echoed back verbatim on the next turn's `chat_history`; a UUID satisfies both. The now-unused `extra` handler arg is dropped.

## Verification

- `pnpm -r build`, `pnpm --filter @vicoop-bridge/client typecheck` ✅
- `pnpm --filter @vicoop-bridge/client test` → **647 pass / 0 fail** ✅ (existing tests inject `callId` via `invokeForTest`, so they're unaffected)
- **Live**: ran the daemon against a real request after the patch. An 8-call session now shows **8 distinct ids, zero collisions, every call paired 1:1 with its result** — vs. the old "2 × 15" collisions.

| | before (bug) | after |
|---|---|---|
| distinct ids | 5 (2,3,4,5,6) | all unique |
| `id="2"` reuse | 15× | — |
| call↔result 1:1 | broken | ✅ |

## Notes

- claude backend only, as scoped. Other backends (`codex.ts` etc.) were not inspected for the same pattern in this PR.
- No regression test added: exercising the real `CallToolRequestSchema` id-minting path needs an MCP-transport-level integration test (current tests bypass it via `invokeForTest`). Happy to add one if wanted.

### DIFF
--- a/.changeset/caller-tool-id-uniqueness.md
+++ b/.changeset/caller-tool-id-uniqueness.md
@@ -1,0 +1,12 @@
+---
+"@vicoop-bridge/client": patch
+---
+
+claude backend: mint a conversation-unique `call_<uuid>` for each
+caller-tool invocation instead of reusing the MCP JSON-RPC request id.
+The request id only counts within a single MCP session, and the bridge
+stands up a fresh MCP server per A2A turn, so it reset every turn —
+every single-call turn emitted `tool_call_id` "2". Once the OpenAI loop
+replayed the accumulated history those duplicate ids collided, breaking
+the call↔result pairing and making the model restate its whole plan
+before each tool call.
--- a/packages/client/src/backends/caller-tools-mcp.ts
+++ b/packages/client/src/backends/caller-tools-mcp.ts
@@ -197,17 +197,22 @@ export async function startCallerToolsMcpServer(
     })),
   }));
 
-  server.setRequestHandler(CallToolRequestSchema, async (req, extra) => {
+  server.setRequestHandler(CallToolRequestSchema, async (req) => {
     const name = req.params.name;
     const argsRaw = req.params.arguments;
-    // MCP's wire-level call id (request id from the JSON-RPC envelope) is
-    // the most stable handle for round-tripping back to the caller. Fall
-    // back to a fresh UUID only if `requestId` is missing — which would
-    // be unusual for a real client but keeps the path total.
-    const callId =
-      typeof extra?.requestId === 'string' || typeof extra?.requestId === 'number'
-        ? String(extra.requestId)
-        : `call_${randomUUID()}`;
+    // Always mint a fresh UUID for the tool_call id. We must NOT reuse the
+    // MCP JSON-RPC request id (`extra.requestId`): it is only monotonic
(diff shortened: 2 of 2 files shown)


longfin:
## Live before/after on the preamble repetition

Ran the same prompt ("Japanese visual novel, blonde heroine at sunset, portrait for smartphone") through the daemon before and after the patch, both reaching a 41-turn `<chat_history>`. Comparing the assistant preambles confirms the fix addresses the user-visible symptom, not just the id collision.

### The stutter, gone

**Before (buggy)** — the same plan restated almost verbatim three times in a row (this is the exact run in the original report):

```
[10] I'll wire the visual novel into the app and add the portrait smartphone styling, then update the docs.
[11] I'll wire the visual novel into the app, add portrait smartphone styling, and update the project docs.
[12] I'll add the portrait smartphone styling, then update the project docs.
```

**After (patched)** — progress advances monotonically; each preamble references completed state instead of re-narrating the whole plan:

```
[ 8] I'll generate the heroine character art and the sunset background in parallel now.
[ 9] I've generated both the heroine and the sunset background. Now I'll register the assets...
[12] Assets are registered. Now I'll build the visual novel engine — story data, then the portrait UI...
[14] I'll wire the visual novel into the app and finish up the documentation.
```

### Metrics

| metric | before (bug) | after |
|---|---|---|
| tool_call distinct ids | **5 / 25** (collide) | **22 / 22** (all unique) |
| `id="2"` reuse | 15× | 0 |
| preambles restating the wire/styling/update-docs plan | **7** | **1** |
| consecutive near-verbatim stutter | `[10][11][12]` present | none |
| up-front planning | none | model called `todowrite` |

### Interpretation

With the call↔result pairing restored, the model can track what it already did, so instead of re-stating the full plan before every tool call it moves forward referencing completed steps ("I've generated both…", "Assets are registered…"). It even did up-front planning (`todowrite`) that the broken linkage had been suppressing.

(The "build the VN" phrase appears 8× in the after-run, but those are distinct forward steps — plan → generate art → register assets → build engine → build UI — each naming the VN as its subject, not the verbatim repetition we were chasing.)

Issue #190, opened 2026-05-15
design: unattended-operation / daemonize strategy for vicoop-client


longfin:
## Context

#187 removed the half-built systemd auto-registration from `install.sh` and the `upgrade` try-restart path (`INSTALL_SKIP_SERVICE` / `INSTALL_SERVICE_SCOPE` env vars, `systemctl try-restart` after bundle swap, related docs) because the always-on supervisor story isn't settled yet. The current supported entrypoint is the foreground run; `screen`/`tmux` remain as ad-hoc workarounds.

#186 surfaced a related problem from a different angle: `nohup vicoop-client ... &` exits immediately with an empty log when invoked from Codex CLI's `exec_command` (non-PTY) environment. Diagnostics narrowed it down:

- The wrapper reaches `exec node`, then Node dies before any `console.log` lands on disk (consistent with SIGKILL — no flush window).
- The same command works fine under PTY + interactive zsh, foreground, `screen`, plain `bash -c` / `zsh -c`, and other non-PTY shells.
- The likely cause is Codex's `exec_command` sending a process-group SIGKILL after the foreground returns from `&`; `nohup` only sets SIGHUP → SIG_IGN, which doesn't protect against SIGKILL.

So we have two pieces of the same puzzle:

1. **Removed**: a Linux-only auto-installed systemd unit (was the existing answer for unattended operation).
2. **Observed**: ad-hoc `nohup &` is brittle in agent-driven exec environments.

A coherent design needs to cover both. Filing this issue so the next attempt isn't another partial implementation.

## Design questions

1. **Supported supervisor matrix per OS** — what do we tell operators (and agents driving the client) to use for unattended operation?
    - Linux: systemd unit. Auto-installed? Operator-installed via a generator subcommand? Doc-only template?
    - macOS: launchd LaunchAgent / LaunchDaemon. Same auto-vs-template question.
    - Other (containers, BSDs, k3s pods, …): foreground + recommended external supervisor (s6, runit, dumb-init, etc.).
2. **Self-daemonize mode** — do we ship a `--daemonize` flag that uses Node's `spawn({ detached: true })` to setsid()-equivalent into a fresh session/process group? This would survive the Codex-style process-group SIGKILL.
    - Pro: works in restrictive agent-exec environments without an external supervisor.
    - Con: conflicts with proper supervisor expectations (systemd `Type=simple` and launchd both want the child NOT to fork). Must be opt-in if we ship it.
[...]

PR #334, opened 2026-06-09
feat(client): unattended operation via `start --detach` + `stop`/`status` (#190 tier 0)

### LINKED ISSUE #186: macOS background client run via nohup exits immediately and screen quit can leave child processes
## Summary

While setting up a local `vicoop-bridge-client` on macOS, foreground execution worked, but a simple `nohup ... &` background run exited immediately with no log output. Running under `screen` worked, but stopping the detached screen session did not fully clean up the child `vicoop-client` / `codex app-server` processes; they had to be killed by PID.

## Environment

- OS: macOS / Darwin
- Install path: `/Users/seongwonmun/vicoop-bridge-client`
- Client version: `0.15.0`
- Backend: `codex`
- Agent id: `f2b79c26-codex-mac`
- Bridge: `https://vicoop-bridge-server.fly.dev`
- `codex --version`: `codex-cli 0.130.0`

## What worked

Foreground run connected successfully:

```sh
"$HOME/vicoop-bridge-client/bin/vicoop-client" --backend codex
```

It printed the expected connection details, including:

```text
[client] connected, sending hello
[client] agentId:    f2b79c26-codex-mac
[client] mention:    @f2b79c26-codex-mac@vicoop-bridge-server.fly.dev
```

The agent card endpoint also returned 200 while the process was running.

## What failed

Attempting to detach with `nohup` exited immediately:

```sh
mkdir -p "$HOME/.vicoop"
nohup "$HOME/vicoop-bridge-client/bin/vicoop-client" --backend codex \
  > "$HOME/.vicoop/vicoop-client.log" 2>&1 &
```

Observed behavior:

- The process disappeared almost immediately.
- `~/.vicoop/vicoop-client.log` stayed empty.
- Running the same command in the foreground worked.

## Workaround used

`screen` was available, so I used:

```sh
screen -dmS vicoop-client "$HOME/vicoop-bridge-client/bin/vicoop-client" --backend codex
```

That kept the client online.

## Shutdown issue

When stopping the daemon, this removed the screen socket:

```sh
screen -S vicoop-client -X quit
```

But related processes were still present afterward:

```text
node -- /Users/seongwonmun/vicoop-bridge-client/bin/../dist/cli.js --backend codex
node /Users/seongwonmun/.npm/bin/codex app-server
.../codex app-server
```

They had to be terminated explicitly by PID.

## Expected behavior

- The docs should recommend a reliable macOS background-running method, or the client should tolerate `nohup`/non-interactive background execution if that is intended to work.
- Stopping the wrapper/session should reliably terminate the backend child process tree, especially for `codex app-server`.

## Notes

This happened during manual setup from `docs/install-client.md`. Since the installer already warns that it targets Linux/Fly.io containers and macOS proceeds best-effort, this may be primarily a documentation gap, but the lingering child process behavior may also be worth handling in the client shutdown path.

longfin:
## What

Adds background/unattended operation for `vicoop-client` without an external supervisor — the **tier-0** half of the [#190](https://github.com/planetarium/vicoop-bridge/issues/190) supervisor design ([decision comment](https://github.com/planetarium/vicoop-bridge/issues/190#issuecomment-4655414354)).

```sh
vicoop-client start --detach --backend openclaw
vicoop-client status        # running (pid …, up …, logs …)   exit 0
vicoop-client stop          # SIGTERM → grace → SIGKILL        exit 0
```

## Why

`nohup vicoop-client … &` exits immediately with an empty log in agent-driven exec sandboxes (Codex `exec_command`, #186): `nohup` only sets `SIGHUP→SIG_IGN`, no defense against the SIGTERM/SIGKILL such environments send to the whole process group when a shell turn ends. `screen -dmS` works but is an extra dependency, and there's **no portable shell one-liner** (macOS ships no `setsid`).

`--detach` re-execs the daemon as a new **session/process-group leader** (`spawn({detached:true})` + `unref()`), which is exactly what survives session/pgrp reaping.

## How

- **Parent writes the pidfile right after fork** (it already holds the child pid) — sidesteps the classic `Type=forking` read-before-write race. A short grace window turns an immediate config/credential failure into a non-zero launch + log pointer instead of a "started" message over a corpse.
- **`stop` / `status` defend against PID reuse**: liveness via `kill(pid,0)` **plus** a best-effort `ps` command-match before signaling/reporting. A recycled or crashed PID is reported `stale` and cleaned up, never killed. `status` uses LSB exit codes (0 running / 3 stopped / 1 stale).
- The detached child owns its pidfile (`VICOOP_DETACHED=1`) and removes it on graceful exit.
- `--detach` / `--log-file` are kept separate from the runtime-config flag surface so they don't conflict with a real supervisor's expectations.

## Scope / non-goals

- **POSIX-only.** `--detach` on win32 errors with a pointer to a Windows service manager (NSSM).
- **Caveat (documented):** `--detach` survives session/pgrp teardown but **not** cgroup/container teardown or reboot. That persistence tier — `service install` generating a launchd LaunchAgent (macOS) + systemd `--user` unit (Linux), **both OSes together** per the issue non-goal — is **tier 1**, a follow-up PR. `--detach` stays the no-supervisor-available fallback, labeled as such.
- `install.sh` stays bundle-staging only; `upgrade` stays supervisor-agnostic.

## Tests

- New `daemon-control.test.ts` — 16 unit tests covering pidfile read/write (+0600 mode), liveness, `inspectDaemon` running/stale/stopped classification, and `stopDaemon` outcomes (probe/kill/wait injected, no real OS processes).
- Full client suite green (**676 passing**); typecheck + build green across all packages.
- Manual smoke (built binary): detach→status(0)→stop(0)→status(3) with pidfile auto-cleanup; immediate-exit detection (exit 1 + log pointer); double-start guard.

Closes #186. Part of #190 (tier 0).


longfin:
Follow-up commit `c051762` closes the two ways a pidfile lifecycle turns into the "reinvent init badly" anti-pattern, so tier-0 isn't just a naive pidfile:

1. **Atomic claim (no double-spawn).** `start --detach` now claims the pidfile with an `O_EXCL` create (`claimPidFile`) instead of check-then-spawn. The winner writes a live, command-matching placeholder (the launcher process) for the start window; racers that lose the create see a coherent "running" record and back off. Smoke-tested: **5 concurrent `start --detach` → exactly one daemon**, the other four refuse.

2. **Principled PID-reuse guard.** Records the OS process **start identity** at spawn (Linux `/proc/<pid>/stat` starttime, macOS `ps -o lstart`) and compares it against the live value in `stop`/`status`. An exact mismatch is authoritative ("recycled PID → not ours"); the earlier `ps` command-substring match is now only the fallback for when start identity isn't available on both sides.

Every `startDetached` failure path releases the claim. +5 unit tests; full client suite green (**681 passing**).

Issue #302, opened 2026-05-28
Backends: read openai-compat/v1 envelope directly, drop decomposition shim


longfin:
Follow-up to #297 (closes oai2a2a#80's envelope contract on the bridge side).

## Background

The 0.4.0 wave of planetarium/oai2a2a#80 / planetarium/vicoop-bridge#297 made the openai-compat/v1 extension's wire fully envelope-based — the canonical request payload is \`metadata[URI].chat_completions_request\` (a verbatim OpenAI Chat Completions request body).

To minimize that PR's scope and test churn, the bridge's 4 backends (claude / codex / vicoop-codex / openclaw) were not rewritten. Instead, \`parseOpenAICompatMetadata\` in \`packages/client/src/backends/openai-compat.ts\` decomposes the envelope back into the existing 4-field \`OpenAICompatMetadata\` projection (\`{ system, tools, tool_choice, chat_history }\`) that the backends already consume. Backends saw no code change.

This shim is transitional. Spec already says the envelope is canonical and agents \"read \`messages[]\` etc. directly off this envelope.\" The decomposition exists only as an in-memory derived view for the backends; it has no wire representation.

## Goal

Migrate each backend to read \`metadata[URI].chat_completions_request\` directly. After the migration, delete the decomposition shim and the \`OpenAICompatMetadata\` interface.

### Backend rewrites (per backend)

For each of **claude.ts**, **codex.ts**, **vicoop-codex.ts**, **openclaw.ts**:

- Replace consumption of \`OpenAICompatMetadata.system\` / \`tools\` / \`tool_choice\` / \`chat_history\` with envelope reads:
  - \`envelope.messages[].filter(m => m.role === 'system' || m.role === 'developer')\` → system prompt assembly
  - \`envelope.tools\` → native tool schema translation
  - \`envelope.tool_choice\` → native tool_choice
  - \`envelope.messages[]\` (excluding trailing user, which arrives via A2A parts) → multi-turn replay
- Use \`envelope.model\` as the dispatch model arg (closes the remaining gap — see oai2a2a#80 thread for context; backends currently use their own CLI defaults).
- Update each backend's tests to cover envelope-direct read paths.

### Parser cleanup (after all 4 backends migrate)

- Delete \`parseEnvelope\`, \`parseLegacyDecomposed\`, \`parseLegacyChatHistory\`, \`collectSystemFromMessages\`, \`chatHistoryFromMessages\`, \`parseHistoryEntry\`.
[...]

PR #349, opened 2026-06-10
fix(client): report actual response model in claude openai-compat envelope (#348)

### LINKED ISSUE #348: Claude backend OpenAI-compatible response model can differ from the actual requested model
## Summary

After upgrading to `vicoop-client 0.32.0` and advertising multiple Claude models via `--claude-supported-models`, requests are routed to the correct Claude model, but the OpenAI-compatible response envelope can report a different top-level `model` value.

This looks like a response-envelope bug rather than a model-routing failure.

## Setup

Service command:

```bash
vicoop-client start \
  --backend claude \
  --cwd /home/vicoop/vicoop-bridge \
  --claude-supported-models claude-opus-4-8,claude-sonnet-4-6,claude-fable-5
```

Agent card advertised models correctly:

```json
[
  { "id": "claude-opus-4-8", "default": true },
  { "id": "claude-sonnet-4-6" },
  { "id": "claude-fable-5" }
]
```

## Reproduction

Send OpenAI-compatible chat requests through the bridge/router with each model:

```json
{
  "model": "claude-sonnet-4-6",
  "messages": [{ "role": "user", "content": "Reply with exactly: router-ok" }],
  "max_tokens": 32,
  "temperature": 0
}
```

Same for:

- `claude-opus-4-8`
- `claude-sonnet-4-6`
- `claude-fable-5`

All requests returned HTTP 200 and `router-ok`.

## Observed behavior

The top-level OpenAI-compatible response `model` field did not always match the requested model. For example, requests to `claude-sonnet-4-6` and `claude-fable-5` returned an envelope with:

```json
{
  "model": "claude-haiku-4-5-20251001"
}
```

However, the local Claude Code JSONL session logs showed the actual Claude CLI execution model was correct:

```text
request claude-opus-4-8   -> message.model = claude-opus-4-8
request claude-sonnet-4-6 -> message.model = claude-sonnet-4-6
request claude-fable-5    -> message.model = claude-fable-5
```

Example JSONL location on the host:

```text
~/.claude/projects/-tmp-vicoop-bridge-claude-oai/*.jsonl
```

Relevant JSONL field:

```json
{
  "type": "assistant",
  "message": {
    "model": "claude-sonnet-4-6",
    "role": "assistant"
  }
}
```

## Expected behavior

The OpenAI-compatible response envelope should report the actual model that handled the request, ideally the model surfaced by Claude Code in the assistant message:

```json
{
  "model": "claude-sonnet-4-6"
}
```

At minimum, it should not report an unrelated/default model when the Claude CLI log confirms that the requested override model was used.

## Why it matters

Consumers and routers often use the response `model` field for accounting, debugging, and validation. If the field reports a default or unrelated Claude model, operators may incorrectly conclude that model override failed even when the backend executed the right model.

longfin:
Fixes #348.

## Problem

After advertising multiple Claude models via `--claude-supported-models`, requests routed to the correct Claude model, but the OpenAI-compatible response envelope's top-level `model` could report a different, unrelated model (e.g. `claude-haiku-4-5-20251001`) — even though the Claude CLI JSONL log confirmed the requested override model actually handled the request.

### Root cause

The envelope's `model` was derived from `result.modelUsage` by picking the entry with the **largest output-token share** (`parseClaudeModelUsageForOpenAICompat`). On a short response (`router-ok`), the actual response model emits very few output tokens, while an internal sub-model Claude Code uses for things like title generation (`claude-haiku-4-5-*`) can out-produce it — so haiku "won" the heuristic and got stamped on the envelope. This is fine for token accounting but wrong as a "which model answered" label.

## Fix

Resolve the response model from ids **claude itself reports**, in order:

1. **The model named on the `assistant` stream event** — the model that actually produced the user-facing turn (the field the issue itself points at).
2. **The `system/init` model** — claude's own report of the model it resolved to run with — used when no assistant event names one (e.g. a `result`-only turn).

The requested `envelope.model` is **deliberately not** used as a fallback: it can be an unresolved routing key — a slug or an A2A card url — which claude drops before `--model` so it runs its own default (#302). Echoing that back as the model would be the same class of bug. Both chosen sources are always real model ids.

The envelope's embedded `usage.model` is stamped with the same resolved id so the spec's "id SHOULD match usage.model" cross-check holds. Token **sums** are untouched — they remain the across-sub-model totals from `modelUsage`. The old largest-output-share model heuristic is removed; `parseClaudeModelUsageForOpenAICompat` now only sums token counts.

## Tests

- New: envelope reports the assistant-turn model, not the modelUsage winner (the #348 repro).
- New: envelope falls back to the `system/init` model on a result-only turn.
- New: envelope never echoes a requested routing slug / A2A card url — reports claude's resolved model instead.
- Updated: `modelUsage` token-sum tests no longer assert a heuristic model label; the helper unit test asserts sums only.

`pnpm -r build`, client `typecheck`, and the full client suite (708 tests) all pass. Includes a client patch changeset.

### DIFF
--- a/.changeset/claude-envelope-model-348.md
+++ b/.changeset/claude-envelope-model-348.md
@@ -1,0 +1,13 @@
+---
+'@vicoop-bridge/client': patch
+---
+
+claude backend: report the actual response model in the OpenAI-compatible
+envelope. The envelope's top-level `model` (and its embedded `usage.model`)
+now resolve from the model named on the assistant turn, falling back to the
+requested model id (already forwarded to claude as `--model`). The previous
+`result.modelUsage` largest-output-share heuristic — which on short responses
+could be dominated by an internal sub-model (e.g. `claude-haiku-4-5-*` used
+for title generation) and mislabel the envelope even when the requested
+override model handled the request — has been removed; `modelUsage` is now
+used only to sum token counts (#348).
--- a/.changeset/claude-envelope-model-348.md
+++ b/.changeset/claude-envelope-model-348.md
@@ -4,10 +4,12 @@
 
 claude backend: report the actual response model in the OpenAI-compatible
 envelope. The envelope's top-level `model` (and its embedded `usage.model`)
-now resolve from the model named on the assistant turn, falling back to the
-requested model id (already forwarded to claude as `--model`). The previous
-`result.modelUsage` largest-output-share heuristic — which on short responses
-could be dominated by an internal sub-model (e.g. `claude-haiku-4-5-*` used
-for title generation) and mislabel the envelope even when the requested
-override model handled the request — has been removed; `modelUsage` is now
-used only to sum token counts (#348).
+now resolve from model ids claude itself reports — the model named on the
+`assistant` turn, falling back to the `system/init` resolved model — instead
+of the `result.modelUsage` largest-output-share heuristic, which on short
+responses could be dominated by an internal sub-model (e.g.
+`claude-haiku-4-5-*` used for title generation) and mislabel the envelope even
+when the requested override model handled the request. The requested
+`envelope.model` is deliberately not used as a fallback, since it may be a
+routing slug or an A2A card url rather than a real model id. `modelUsage` is
+now used only to sum token counts (#348).
--- a/packages/client/src/backends/claude.test.ts
+++ b/packages/client/src/backends/claude.test.ts
@@ -3278,10 +3278,12 @@ test('result.modelUsage is summed across models and forwarded as openai-compat u
   assert.equal(payload.usage.total_tokens, 25287 + 21);
   // cached_tokens mirrors Σ_M cacheRead (lossless): 0 + 18029.
   assert.deepEqual(payload.usage.prompt_tokens_details, { cached_tokens: 18029 });
-  // `model` reports the entry with the largest output share — opus had 8 vs
-  // haiku's 13, so haiku wins by raw output_tokens count. The contract is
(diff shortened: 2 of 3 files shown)


longfin:
## Verified against a real Claude Code capture

Ran the actual `claude` CLI (2.1.170) in the issue's scenario — a short reply with an explicit model — and captured the stream-json, then replayed those exact lines through the backend's `handle()`.

**Command**
```
claude -p "Reply with exactly: router-ok" --output-format stream-json --verbose \
  --model claude-sonnet-4-6 --max-turns 1
```

**What the real stream actually carries**

| stream-json field | real value |
| --- | --- |
| `system/init.model` | `claude-sonnet-4-6` ← new fallback source |
| `assistant.message.model` | `claude-sonnet-4-6` ← new primary source |
| `result.modelUsage` output tokens | **haiku = 11, sonnet = 6** |

So this is a genuine repro: the haiku sub-model out-produced the answering model (11 > 6 output tokens), meaning the old largest-output-share heuristic would have stamped `claude-haiku-4-5-20251001` on the envelope.

**Replaying the captured JSONL through `handle()` (post-fix)**
```
[VERIFY-348] envelope.model = "claude-sonnet-4-6"
[VERIFY-348] envelope.usage.model = "claude-sonnet-4-6"
✔ real claude capture → envelope.model = answering model
```

The envelope now reports the model that actually answered, and `usage.model` stays consistent — not the haiku winner. Full client suite (708 tests) green.

PR #354, opened 2026-06-10
feat(server): add Sentry (@sentry/hono) monitoring

ost006:
## What

Adds Sentry error + performance monitoring to the Hono API server (`@vicoop-bridge/server`), reporting to the shared **`vicoop-router-monitoring`** Sentry project (org `vicoop-router`) — the same project `a2x-internal-router` reports to.

## Changes
- **`packages/server/src/instrument.ts`** (new): `Sentry.init()` via `@sentry/hono/node`. Tagged `service=vicoop-bridge-server`.
- **`packages/server/src/cli.ts`**: import `./instrument.js` first, before any other module, so Sentry initializes before app code loads.
- **`packages/server/src/http.tsx`**: `app.use(sentry(app))` registered as the first middleware so it wraps the full request lifecycle and auto-captures errors from `app.onError`.
- **`packages/server/fly.toml`**: `SENTRY_ENABLED="true"` + `SENTRY_ENVIRONMENT="production"` so monitoring activates on deploy.

## DSN / secrets
- **No hardcoded DSN.** Read from the **`SENTRY_DSN`** env var, set as a **Fly secret** (`fly secrets set SENTRY_DSN=…`). If unset, the SDK initializes disabled and sends nothing.
- Gated on `SENTRY_ENABLED` (set in fly.toml) so local `tsx` runs stay silent without forcing `NODE_ENV`.

## Notes
- This is a **Hono** app (`@hono/node-server`), so it uses `@sentry/hono` — not the generic Node or Express SDK. (Express only backs the separate PostGraphile process.)
- **Single shared Sentry project**: this service is distinguished from `a2x-internal-router` by the `service` tag.
- `pnpm --filter @vicoop-bridge/server typecheck` passes.

### DIFF
--- a/packages/server/fly.toml
+++ b/packages/server/fly.toml
@@ -10,6 +10,10 @@ primary_region = "nrt"
 [env]
   PORT = "8787"
   POSTGRAPHILE_PORT = "5433"
+  # Sentry → shared "vicoop-router-monitoring" project. Enables the SDK in this
+  # deployment (see packages/server/src/instrument.ts; DSN is a baked default).
+  SENTRY_ENABLED = "true"
+  SENTRY_ENVIRONMENT = "production"
 
 [http_service]
   internal_port = 8787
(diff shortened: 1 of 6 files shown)


longfin (approved):
LGTM

PR #371, opened 2026-06-17
fix(client): strip cache_control from replayed openai-compat history

longfin:
## What

On the **openai-compat** path, strip Anthropic-style `cache_control` markers from the replayed `<chat_history>` block before rendering it into the prompt.

## Why

While investigating a Claude-backend usage spike on the provider host, the breakdown of one busy hour (88 sessions, ~25.4M tokens) showed **cache _writes_ were 58% of all tokens** — the dominant cost — even though it's an OpenAI-compatible (stateless) path where every request resends the full growing history, so consecutive requests share a ~99.8%-identical prefix that *should* be cache-**read**.

Per-turn analysis showed every session's first model turn read a **constant 42,327 tokens** (just Claude Code's static system+tools+`CLAUDE.md` prefix) and **wrote** a growing 25k→121k (the whole history). The 100k+ shared history was re-written every turn, never read.

Diffing two consecutive history blocks pinned the cause to a byte mutation **in the middle** of the otherwise-identical array (char 24,841 of 59,726 — ~35k chars before the end):

```diff
           "name": "mcp___vb-caller-tools__read",
           "arguments": "{\"filePath\":\"/home/project/.env\"}"
-        },
-        "cache_control": { "type": "ephemeral" }
+        }
       }
```

An Anthropic `cache_control: {"type":"ephemeral"}` marker rides inbound on `messages[]` (the gateway / calling agent places one ephemeral breakpoint on a recent turn). It **moves to the latest turn every request** and is removed from the previously-marked entry. Because `formatChatHistory` serialized the history verbatim (`JSON.stringify(history, null, 2)`), the marker's movement re-encoded a mid-array entry on every follow-up turn → the Anthropic prompt-cache prefix match broke at the marker's position → the entire replayed transcript after it was re-written to cache (1.25×) instead of read (0.1×).

Compounding it, the openai-compat path intentionally disables claude session reuse (`claude.ts`), so every request is a cold spawn with no warm in-process cache to absorb the miss.

## Fix

Deep-strip every `cache_control` key from the history before rendering. The markers are an API-transport hint, not conversation content, and were never part of the documented history shape the model is told to expect (`buildOpenAICompat*SystemPrompt`). Removing them makes the rendered block a stable append-only prefix, so the replayed history stays cache-readable across turns. The portion that was re-written each turn moves from cache-write (1.25×) to cache-read (0.1×) — ~12× cheaper on that span, and a proportional drop in subscription-quota pressure.

Strip happens at render time in `formatChatHistory` (shared by the claude / openclaw text-prepend backends). codex's native `historyToInjectItems` path is untouched.

## Testing

- `pnpm -r build`, `pnpm --filter @vicoop-bridge/client typecheck` — clean
- `pnpm --filter @vicoop-bridge/client test` — **721 pass / 0 fail**
- New regression tests cover: markers stripped wherever they ride (message-level + nested in `tool_calls`), semantic content preserved, **a moved marker renders byte-identical** (the caching invariant), and the caller's history is not mutated.

### DIFF
--- a/.changeset/openai-compat-history-cache-control.md
+++ b/.changeset/openai-compat-history-cache-control.md
@@ -1,0 +1,18 @@
+---
+'@vicoop-bridge/client': patch
+---
+
+Strip `cache_control` markers from the replayed `<chat_history>` block on the
+openai-compat path. Anthropic-style `cache_control: {"type":"ephemeral"}`
+markers ride inbound on `messages[]` (the gateway / calling agent places one
+ephemeral breakpoint on a recent turn). `formatChatHistory` rendered the history
+verbatim, so the marker — which moves to the latest turn every request — mutated
+an entry in the middle of the otherwise byte-identical history on each follow-up
+turn. That broke the Anthropic prompt-cache prefix match at the marker's
+position, forcing the entire replayed transcript after it to be re-written to
+cache (1.25×) instead of read (0.1×) on every turn. Since openai-compat
+conversations resend the full growing history each turn, this dominated token
+usage on long sessions. The markers are an API-transport hint, not conversation
+content, and were never part of the history shape the model is told to expect —
+they are now deep-stripped before rendering so the block is a stable
+append-only prefix and the replayed history stays cache-readable across turns.
--- a/packages/client/src/backends/openai-compat.test.ts
+++ b/packages/client/src/backends/openai-compat.test.ts
@@ -442,6 +442,66 @@ test('formatChatHistory: returns empty when the history is empty', () => {
   assert.equal(formatChatHistory([]), '');
 });
 
+test('formatChatHistory: strips cache_control markers wherever they ride', () => {
+  // Anthropic-style cache_control markers ride inbound on messages[] (gateway /
+  // caller places an ephemeral breakpoint). They are transport hints, not
+  // conversation content — rendering them verbatim leaks API noise into the
+  // prompt and, because the marker moves each turn, thrashes the prompt cache.
+  const rendered = formatChatHistory([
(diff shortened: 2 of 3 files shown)


longfin:
Closing — the central claim here is **refuted by direct measurement** on the provider host.

I ran a real `claude -p` A/B (opus-4-8, 1h cache, clean cwd, ~59k-token prefix), reading actual `usage` tokens:

| call | in | cache_write | cache_read |
|---|--:|--:|--:|
| 1 warm | 2488 | 43,638 | 15,328 |
| 2 repeat (byte-identical) | 2 | 2,486 | **58,966** |
| 3 append-only (no marker) | 2488 | 43,663 | **15,328** |
| 4 append + mid-prefix marker | 2488 | 43,690 | **15,328** |

Findings:
1. Cross-spawn caching works **only for byte-identical full-prompt repeats** (call 2 reads the whole body).
2. **Append-only already defeats it** (call 3): Claude Code in `-p` mode anchors the body's cache breakpoint at the prompt end, so any tail change (one appended turn) makes it unreadable — only the static system/tools prefix stays cached cross-turn. This matches the production "flat 42k read + full body re-write every turn".
3. **The marker is irrelevant** (call 3 vs call 4 are byte-for-byte identical in cache behavior). Stripping it changes nothing.

So this PR does **not** reduce cache-write cost — the spike is structural (stateless cold spawn re-sending the growing history with no history-boundary breakpoint), not the moving `cache_control` marker. The real levers are warm-session reuse or a breakpoint at the stable history boundary, both out of scope for this change. Closing rather than shipping a fix whose premise didn't hold up.

PR #374, opened 2026-06-17
feat: canonical BridgeUsage shape for the usage API (+ Claude remaining usage)

ost006:
## What

Two things, stacked:

1. **Claude remaining usage** — the bridge usage API (`GET /admin-api/agents/:id/usage`) now works for the **claude** backend (was hardcoded to `vicoop-codex`).
2. **Canonical shape** — both backends now report ONE backend-agnostic shape, `BridgeUsage`, so a single consumer can read remaining quota regardless of backend.

> The endpoint has **no consumers today** (admin-ui talks to an LLM agent over A2A; a2x-router uses its own DB) — so this is a clean contract, not a migration.

## Why a canonical shape

The two backends' raw payloads disagreed on everything that matters:

| concept | vicoop-codex (raw) | claude oauth (raw) | claude rate_limit_event (raw) |
|---|---|---|---|
| usage | `remaining_percent: 80` | `utilization: 21` (0–100) | `utilization: 0.93` (**0–1**) |
| reset | `reset_at` (epoch s) | `resets_at` (ISO) | `resetsAt` (epoch s) |
| structure | `{accounts:[…]}` | single snapshot | single window |

Normalizing fixes the used-vs-remaining / units / time-format muddle before it reaches any consumer.

## Canonical shape (`@vicoop-bridge/protocol`)

```ts
BridgeUsage {
  backend: string;            // 'claude' | 'vicoop-codex'
  source: string;             // 'oauth' | 'serve' | 'rate_limit_event' | 'none'
  fetchedAt: string;          // ISO 8601
  accounts: {
    id: string; label?: string; plan?: string;
    windows: { id; label; usedPercent; resetsAt; severity }[];
    spend?: { usedMinor; limitMinor; currency; usedPercent; resetsAt };
    note?: string;
  }[];
  note?: string;
  raw?: unknown;              // verbatim upstream payload
}
```

Fixed conventions:
- **`usedPercent`**: 0–100, percent **used** (remaining = `100 − usedPercent`); severity derived uniformly (≥90 critical, ≥75 warning).
- **`resetsAt`**: ISO 8601 absolute, or `null`.
- **`accounts`**: always an array (codex is multi-account, claude has one).
- Monetary overage (claude `extra_usage`) → separate **`spend`** block, not mixed into percent windows — fixes the "overage 93% shown as a quota window" confusion.
- **`raw`**: verbatim upstream payload preserved → no info loss.

## How

- **protocol**: add `BridgeUsage`/`UsageAccount`/`UsageWindow`/`UsageSpend` zod schemas + types. Wire field stays `z.unknown()` (forward-compat / graceful degrade); backends produce `BridgeUsage`.
- **claude backend** (`claude-usage.ts`): read the Claude Code OAuth token cross-platform (macOS Keychain / `~/.claude/.credentials.json` on Linux & Windows), call `api/oauth/usage`, map windows (`five_hour → session_5h`, `seven_day → weekly`, `seven_day_sonnet → weekly_sonnet`, …) + `extra_usage → spend`. 5-min cache + once-per-TTL throttle for the endpoint's self-429. Fallback: the captured `rate_limit_event` window (`utilization` 0–1 → 0–100), previously parsed-and-discarded at `claude.ts`. Never throws — degrades with a `note`.
- **vicoop-codex backend**: normalize serve `/usage` `{accounts:[…]}` into the same shape (`primary`/`secondary` → `session_5h`/`weekly` by window length; `remaining_percent` → `usedPercent`) instead of forwarding verbatim.
- **server**: open the gate from `vicoop-codex`-only to a `['vicoop-codex','claude']` whitelist.
- shared unit/format helpers in `usage-normalize.ts`.

## Example output (both backends, same shape)

```jsonc
// claude
{ "backend": "claude", "source": "oauth", "fetchedAt": "…",
  "accounts": [{ "id": "default", "plan": "team",
    "windows": [
      { "id": "session_5h",    "usedPercent": 21, "resetsAt": "2026-06-17T09:19:59Z", "severity": "ok" },
      { "id": "weekly",        "usedPercent": 35, "resetsAt": "2026-06-22T01:59:59Z", "severity": "ok" },
      { "id": "weekly_sonnet", "usedPercent": 12, "resetsAt": "2026-06-22T01:59:59Z", "severity": "ok" }
    ],
    "spend": { "usedMinor": 9290, "limitMinor": 10000, "currency": "SGD", "usedPercent": 93, "resetsAt": "2026-07-01T00:00:00Z" }
  }], "raw": { "...": "verbatim /api/oauth/usage" } }

// vicoop-codex — identical structure
{ "backend": "vicoop-codex", "source": "serve", "fetchedAt": "…",
  "accounts": [{ "id": "k1", "label": "you@example.com", "plan": "plus",
    "windows": [
      { "id": "session_5h", "usedPercent": 41, "resetsAt": "…", "severity": "ok" },
      { "id": "weekly",     "usedPercent": 68, "resetsAt": "…", "severity": "ok" }
    ]
  }], "raw": { "accounts": ["...verbatim serve /usage..."] } }
```

## Tests

- `claude-usage.test.ts` — cred read on macOS/Linux/Windows, fetch ok/non-2xx, and the provider mapping to canonical (oauth windows + spend, overage-reset threading, 0–1→0–100 fallback, cache/throttle, expired/no-token/HTTP-error degrade).
- `vicoop-codex.test.ts` — serve `/usage` → canonical normalization (+ `raw` preserved).
- Client **746/746 pass**, server **0 fail**, `pnpm -r typecheck` clean.

## Notes

- ⚠️ **ToS** (unchanged from before): calling `api/oauth/usage` with the subscription OAuth token is programmatic access outside the official client (Consumer Terms §3 carves out only API-key access). Flagged for an explicit operator decision before merge; the codex path and the `rate_limit_event` fallback are unaffected by this.


ost006:
### Update — spec-alignment hardening (commit `ee8838b`)

Compared the claude `oauth/usage` read path against an internal reference spec for the same task (a standalone Claude usage monitor) and pulled in its robustness features. Applied items **1–6**; skipped **7** (a second `oauth/profile` call just for account email/uuid — display-identity parity isn't worth it).

| # | Gap vs. spec | Now |
|---|---|---|
| 1 | hardcoded `~/.claude` | honor **`$CLAUDE_CONFIG_DIR`** for the creds file (keychain stays primary on macOS) |
| 2 | generic node UA | send **`User-Agent: claude-code/<version>`** (discovered from `claude --version`, cached, fallback constant) + `Content-Type`, mirroring the official client |
| 3 | 429 = fixed 5-min TTL | parse **`Retry-After`** and back off exactly that long |
| 4 | drop to stream window on any failure | serve the **last successful snapshot** (stale, annotated via `note`) on a transient failure |
| 5 | no refresh | on expiry/401, **delegate a refresh to the CLI** (`claude update`, injectable), re-read creds, retry once |
| 6 | time-based throttle only | **retry-storm guard**: don't re-send a token that got a 401 until its value actually rotates |

Where this implementation already led the spec (kept): **macOS Keychain** read (the spec is file-only → would find no token on a standard macOS install), and the **canonical multi-backend normalization**.

⚠️ Note on item 5: the default refresh runs `claude update`, which can also update the CLI binary — it's injectable/overridable for operators who pin the binary. Flagged for review.

All seams (cred read, fetch, CLI version, refresh) are injectable; tests never spawn a subprocess. **Client 753/753, server 0 fail, typecheck clean.**

ost006:
### Update — dropped the `rate_limit_event` usage fallback (commit `96ed3d0`)

Live-checked against a real account. The stream `rate_limit_event` only ever carries the **single most-constrained window**, and with extra-usage near its cap that's `overage` — so the fallback was returning the wrong thing:

| source | returned |
|---|---|
| `rate_limit_event` (old fallback) | `overage` **only, 93%** — zero subscription windows |
| `oauth` (real) | session **38%** · weekly **44%** · sonnet **14%** · overage 92.9% |

Presenting "overage 93%" as the usage hid the subscription windows the caller actually wants — misleading, not a useful degraded view.

**Change:** `usage()` no longer synthesises a window from `rate_limit_event`. When the oauth read is unavailable and there's no cached snapshot, it returns an explicit `source: 'none'` + note. The event is still captured — but **only** to enrich `spend.resetsAt` (the monthly overage reset the oauth `extra_usage` block omits). Removed the now-unused `rateLimitToWindow()`.

Fallback chain: **stale oauth snapshot → `none`** (was: stale → rate_limit_event → none). Client 753/753, typecheck clean.

Issue #375, opened 2026-06-18
feat(client): forward vicoop-codex reasoning + emit liveness heartbeat


longfin:
Part of **planetarium/a2x-internal-router#95** (long reasoning turns false-failover). **Step 4 — must land AFTER oai2a2a codec + router are deployed** (a new bridge against an old codec leaks reasoning into the answer).

## Goal

Make `vicoop-codex` turns observably alive to the router so healthy long-reasoning turns are never false-failed-over, and stream the model's thinking to the caller.

### (a) Reasoning forwarding
- [ ] `ChatCompletionChunk` delta type: add `reasoning_content?: string | null`.
- [ ] `applyChunk`: when `delta.reasoning_content` is present, call a new `onReasoningDelta(text)` (mirror of `onContentDelta`).
- [ ] In the streaming caller, `onReasoningDelta` emits a **reasoning-channel** `task.artifact`: distinct `artifactId`, `artifact.metadata[<ext-uri>] = { channel: "reasoning" }`, `append:true`, `lastChunk:false`. (Requires serve emitting `reasoning_content` — planetarium/vicoop-codex-cli.)

### (b) Liveness heartbeat
- [ ] While streaming, run a ~10s timer that emits a `working` status update with `metadata[<ext-uri>] = { heartbeat: true }` whenever no content/reasoning byte has arrived in the interval. Reset on every delta; clear on completion/fail/abort.
- [ ] This is the liveness floor for reasoning-summary gaps (≤~37s, see #95), tool-call-only turns, and any silent stretch.

### Safety
- [ ] Gate reasoning emission behind a flag / agent-card param so it only turns ON after the codec+router are deployed (#95 rollout order).
- [ ] Consider lifting the heartbeat into the shared backend task loop so `claude`/`openclaw` benefit too (follow-up; vicoop-codex first).

ext-uri = `https://github.com/planetarium/oai2a2a/extensions/openai-compat/v1`

## Tests — `vicoop-codex.test.ts`

- [ ] `reasoning_content` delta → reasoning-channel artifact (separate id + metadata), NOT a normal content artifact.
- [ ] Heartbeat timer (fake timers): emits `working{heartbeat}` after silence; resets on delta; stops at completion.
- [ ] Existing content/tool-call/usage behavior unchanged.

## Release
- [ ] changeset → client release.

PR #377, opened 2026-06-18
feat(client): forward Claude extended-thinking on openai-compat reasoning channel

### LINKED ISSUE #376: feat(client): forward claude extended-thinking as reasoning_content + share heartbeat
Part of **planetarium/a2x-internal-router#95** (long reasoning turns false-failover). Sibling of **#375** (vicoop-codex). Same wire contract, same rollout-order gate — but the `claude` backend reaches the same outcome with **no upstream CLI change**, because the reasoning is already on the wire.

## Why this is a separate issue

#375 covers `vicoop-codex`, where the reasoning genuinely does not reach the wire: `codex-cli serve` drops all `response.reasoning_*` events and never requests `reasoning.summary`, so vicoop-codex-cli#25 must ship first.

`claude` is different: the backend already spawns with `--output-format stream-json --verbose --include-partial-messages` (`packages/client/src/backends/claude.ts:1899-1920`), which mirrors the Anthropic Messages streaming API onto the wire. When extended thinking is on, thinking arrives as `content_block_start{type:"thinking"}` → `content_block_delta{type:"thinking_delta", thinking:"…"}` → `signature_delta` / `redacted_thinking`. We already receive these frames — `extractClaudeStreamTextDelta` (`claude.ts:619-633`) just filters them out (correctly, so thinking never leaks into the answer). So the data is sitting right where we need it; the bridge can complete this alone.

## Pre-req unique to claude: thinking must be ON

There is currently **no thinking enablement** in the claude argv (no `MAX_THINKING_TOKENS`, no thinking config — grep-confirmed). Claude Code's `-p` headless mode defaults extended thinking off, so there is no reasoning on the wire to forward yet. **Step 0** is enabling thinking (env / budget / interleaved-thinking beta) — the analogue of "serve must emit reasoning" for vicoop-codex. Without it the rest is a no-op.

### (a) Reasoning forwarding
- [ ] Enable extended thinking on the spawn (decide budget; gate behind the same flag/agent-card param as #375).
- [ ] Add `extractClaudeStreamReasoningDelta(evt)` — sibling of `extractClaudeStreamTextDelta`, pulls `delta.thinking` from `content_block_delta{type:"thinking_delta"}`.
- [ ] `redacted_thinking`: do **not** forward content (encrypted) — skip or emit a placeholder only.
- [ ] In the `stream_event` branch (`claude.ts:2242-2249`), route reasoning deltas to a new `emitReasoningArtifact` (mirror of `emitAssistantArtifact`, `claude.ts:2094`): distinct `artifactId`, `artifact.metadata[ext-uri] = { channel: "reasoning" }`, `append:true`, `lastChunk:false`. Same wire contract as #95.

### (b) Liveness heartbeat (shared)
- [ ] #375 safety note already calls for lifting the heartbeat into the **shared backend task loop** so `claude`/`openclaw` benefit too. This is the higher-value, reasoning-independent half: claude has its own silent stretches (long tool/MCP runs) that trip the router's first-content/STALL windows.
- [ ] Natural hook: claude already has timing marks (`recorder.mark('firstAssistant')`, `claude.ts:2245,2253`) to reset the heartbeat timer on each delta.
- [ ] Coordinate with #375 so the heartbeat is implemented once, not twice.

### Safety / rollout
- [ ] Same back-compat gate as #95 step order: a new bridge emitting reasoning against an old codec leaks thinking into the answer → codec (oai2a2a#103) + router (#95) must deploy first.

ext-uri = `https://github.com/planetarium/oai2a2a/extensions/openai-compat/v1`

## Tests — `claude.test.ts`
- [ ] `thinking_delta` → reasoning-channel artifact (separate id + metadata), NOT a `claude-message` content artifact.
- [ ] `redacted_thinking` content is not forwarded.
- [ ] Heartbeat after silence; resets on delta; stops at completion (shared-loop tests may live wherever the heartbeat lands).
- [ ] Existing text / tool-call / usage behavior unchanged.

## Release
- [ ] changeset → client release.

## Related
- #375 (vicoop-codex sibling) · planetarium/a2x-internal-router#95 · planetarium/oai2a2a#103

longfin:
Closes #376. Part of planetarium/a2x-internal-router#95 (step 4 — bridge side).

## What

Forwards Claude's extended-thinking to callers as a live **`reasoning` channel**, so the a2x-internal-router stops false-failing-over long silent reasoning turns (the `streaming response stalled … no bytes for 60000ms` → `TIMEOUT` symptom in #95).

Unlike `vicoop-codex` (where `codex-cli serve` drops reasoning on the wire and needs an upstream fix), Claude's thinking already arrives on the wire — the claude backend spawns `--output-format stream-json --include-partial-messages`, so `thinking_delta` frames are already received and merely filtered out today. This PR stops dropping them.

## How

- **`extractClaudeStreamReasoningDelta`** — sibling of the text extractor; pulls `thinking_delta` / opening `thinking` blocks. `redacted_thinking` (encrypted blob) is never forwarded.
- **`emitReasoningArtifact`** — distinct `artifactId`, `extensions: [openai-compat/v1]`, `metadata[openai-compat/v1] = { channel: "reasoning" }`, `append:true`, `lastChunk:false`. Kept separate from the `claude-message` answer artifact so thinking never co-mingles with the answer. The oai2a2a codec maps it to `delta.reasoning_content`.
- **Thinking enablement** — when on, openai-compat spawns get `MAX_THINKING_TOKENS` injected (Claude Code's `-p` headless mode is otherwise silent). Operator's own `MAX_THINKING_TOKENS` export is not clobbered.

## Flag (ON by default)

Configurable flag + config, no env var:

- CLI: `--no-claude-reasoning`
- config.json: `backends.claude.reasoning: false`

⚠️ **Rollout (#95):** ON by default means an operator on a codec **older than oai2a2a 0.6.0** would have thinking folded into the answer (the old codec doesn't understand the channel marker). Disable via the flag/config until the codec+router are deployed.

## Out of scope

The **heartbeat liveness marker** half of #95/#376 (`metadata = { heartbeat: true }` on the `working` status) is intentionally deferred: claude already emits a 30s idle `working` heartbeat, but tagging it needs a `metadata` field on `TaskStatus` (protocol) + the server's A2A `TaskStatusUpdateEvent.metadata` mapping — a cross-package change that lands with the shared-loop heartbeat follow-up noted in #375.

## Tests

- backend: default-on forwarding, `claudeReasoning:false` drops entirely (no answer leak), redacted-thinking not forwarded, `MAX_THINKING_TOKENS` injection
- cli-args: default ON, `--no-claude-reasoning` off, `backends.claude.reasoning:false` off, flag>config precedence, non-claude backend hard error

`pnpm -r typecheck` + client suite (741 tests) green. Changeset included (client minor).

### DIFF
--- a/.changeset/claude-reasoning-channel.md
+++ b/.changeset/claude-reasoning-channel.md
@@ -12,4 +12,6 @@ reasoning turn as alive instead of false-failing-over it
 (planetarium/a2x-internal-router#95, #376). ON by default; disable with
 `--no-claude-reasoning` or `backends.claude.reasoning: false` when the deployed
 oai2a2a codec predates 0.6.0 and can't yet interpret the channel marker.
-Redacted-thinking blocks are never forwarded.
+Redacted-thinking blocks are never forwarded. The injected thinking budget
+(default 8000) is configurable via `--claude-thinking-budget` /
+`backends.claude.thinking_budget`.
--- a/.changeset/claude-reasoning-channel.md
+++ b/.changeset/claude-reasoning-channel.md
@@ -1,0 +1,15 @@
+---
+'@vicoop-bridge/client': minor
+---
+
+feat(client): forward Claude extended-thinking on an openai-compat/v1
+`reasoning` channel. The claude backend surfaces `thinking_delta` stream events
+as a dedicated `claude-reasoning` artifact carrying
+`metadata[openai-compat/v1] = { channel: "reasoning" }`, and injects a
+`MAX_THINKING_TOKENS` budget on openai-compat spawns so Claude Code emits
+thinking on the wire. This lets the a2x-internal-router treat a long silent
+reasoning turn as alive instead of false-failing-over it
+(planetarium/a2x-internal-router#95, #376). ON by default; disable with
+`--no-claude-reasoning` or `backends.claude.reasoning: false` when the deployed
+oai2a2a codec predates 0.6.0 and can't yet interpret the channel marker.
+Redacted-thinking blocks are never forwarded.
--- a/packages/client/src/backends/claude.test.ts
+++ b/packages/client/src/backends/claude.test.ts
@@ -477,6 +477,40 @@ test('injects MAX_THINKING_TOKENS on openai-compat spawns when reasoning is on',
   assert.equal(fake.lastChild()?.env?.MAX_THINKING_TOKENS, '8000');
 });
 
+test('claudeThinkingBudget overrides the injected MAX_THINKING_TOKENS', async () => {
+  const fake = scriptedSpawn({
+    lines: [
+      JSON.stringify({ type: 'system', subtype: 'init', session_id: 'sid' }),
+      JSON.stringify({ type: 'result', subtype: 'success', result: 'ok' }),
+    ],
+    exitCode: 0,
+  });
+
+  const backend = createClaudeBackend({
+    spawn: fake.spawn,
+    claudeReasoning: true,
+    claudeThinkingBudget: 12000,
+  });
+  const { emit } = collect();
+  const envelopeTask: TaskAssignFrame = {
+    ...assign('hello'),
+    message: {
+      role: 'user',
+      messageId: 'm1',
+      parts: [{ kind: 'text', text: 'hello' }],
+      metadata: {
+        [OPENAI_COMPAT_EXTENSION_URI]: {
+          chat_completions_request: { model: 'gpt', messages: [{ role: 'user', content: 'hello' }] },
+        },
+      },
+    },
+    requestedExtensions: [OPENAI_COMPAT_EXTENSION_URI],
+  };
+  await backend.handle(envelopeTask, emit, NEVER);
+
+  assert.equal(fake.lastChild()?.env?.MAX_THINKING_TOKENS, '12000');
+});
+
 test('falls back to result artifact when streaming produced nothing', async () => {
   const fake = scriptedSpawn({
     lines: [
--- a/packages/client/src/backends/claude.test.ts
+++ b/packages/client/src/backends/claude.test.ts
@@ -312,6 +312,171 @@ test('streams Claude stream_event text deltas as append chunks', async () => {
   assert.deepEqual(artifacts.map((a) => a.append), [true, true]);
 });
 
+test('forwards thinking_delta on a reasoning channel when claudeReasoning is on', async () => {
+  const fake = scriptedSpawn({
+    lines: [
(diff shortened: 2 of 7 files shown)


longfin:
## Local validation (wire-level)

Unit tests feed synthetic `thinking_delta` frames, so they can't verify the upstream assumption this PR rests on: **does real Claude Code actually emit thinking on the wire under our spawn flags?** Verified directly against the `claude` binary (no server/DB needed):

```
claude 2.1.181 (Claude Code)
MAX_THINKING_TOKENS=8000 claude -p "<hard prompt>" \
  --output-format stream-json --verbose --include-partial-messages
```

| check | result |
|---|---|
| Thinking emitted on the wire under our production flags + `MAX_THINKING_TOKENS` | ✅ yes |
| `content_block_start` → `block.type === "thinking"`, `block.thinking` (string) | ✅ matches extractor (`keys: [type, thinking, signature]`) |
| `content_block_delta` → `delta.type === "thinking_delta"`, `delta.thinking` (string) | ✅ exact match |
| `signature_delta` ignored (crypto signature, not content) | ✅ not picked up |
| trivial prompt (`17 * 24`) → no thinking block | ✅ no spurious reasoning channel |

Observed `thinking_delta` frame shape:

```json
{ "type": "content_block_delta", "index": 0,
  "delta": { "type": "thinking_delta", "thinking": "…", "estimated_tokens": 50 } }
```

### Operational findings
- **`MAX_THINKING_TOKENS` is required** — without it, no thinking reaches the wire at all (confirms why the backend injects it). Too low a budget on an easy prompt also yields nothing; `8000` (the default here) on a non-trivial prompt produced thinking reliably.
- **Thinking is model/difficulty dependent** — trivial turns emit no thinking block, so default-ON produces no reasoning channel (and no overhead) on easy turns; no spurious artifacts.

### Scope confirmed (kept as-is)
`MAX_THINKING_TOKENS` auto-injection stays **openai-compat (envelope) spawns only** — aligned with #95, which is entirely the openai-compat → router path. Plain A2A claude tasks only emit a reasoning channel if the operator has set `MAX_THINKING_TOKENS` globally themselves.

Tiers: Tier 1 (unit, 741 green) ✅ · Tier 2 (real-claude wire check) ✅ · Tier 3 (full daemon E2E: server + DB + codec `reasoning_content` mapping) not run.

longfin:
Superseded by #379, which consolidates claude + vicoop-codex reasoning **and** the shared liveness heartbeat (including the vicoop-codex heartbeat wiring that fell in the gap between the separate branches). Closing in favor of the unified PR.

PR #392, opened 2026-06-23
fix(server): single-flight schema-tools introspection (#367)

### LINKED ISSUE #367: server(schema-tools): concurrent first calls duplicate introspection (thundering herd)
## 요약 (심각도: 🟡 Low)

스키마 툴 캐시가 비어 있을 때 동시 첫 요청이 들어오면 introspection GraphQL 쿼리와 툴 빌드가 중복 실행된다. 멱등이라 정합성 문제는 없고 startup 1회성 낭비 수준.

## 근거

`packages/server/src/schema-tools.ts:651-652`:
```ts
async function getOrBuildSchema(): Promise<CachedSchema> {
  if (cachedSchema) return cachedSchema;
  const result = await executeGraphQL(INTROSPECTION_QUERY);  // await 경계
  ...
  cachedSchema = {...};
  return cachedSchema;
}
```

`packages/server/src/schema-tools.ts:748-752`:
```ts
export async function getSchemaTools() {
  if (cachedTools) return cachedTools;
  const schema = await getOrBuildSchema();
  ...
  cachedTools = ...;
}
```

`cachedSchema`/`cachedTools`가 null인 동안 동시 호출이 모두 `await` 구간을 통과하면 introspection을 N번 실행하고 마지막 결과로 덮어쓴다.

## 영향

- 정합성 문제 없음(멱등). 부팅 직후 동시 admin 요청 시 PostGraphile introspection 중복 호출 정도.

## 제안

in-flight Promise를 캐싱하는 패턴으로 single-flight 보장:
```ts
let schemaPromise: Promise<CachedSchema> | null = null;
function getOrBuildSchema() {
  if (cachedSchema) return Promise.resolve(cachedSchema);
  if (!schemaPromise) schemaPromise = buildSchema().then(s => (cachedSchema = s));
  return schemaPromise;
}
```

longfin:
## What changed

`schema-tools` caches the introspected PostGraphile schema (`cachedSchema`) and the built AI tool set (`cachedTools`) in module-level variables. Both used a check-then-`await`-then-set pattern, so while the cache was `null` any concurrent callers all passed the guard, each ran a full introspection + tool build, and the last result overwrote the cache (thundering herd, #367).

This makes both caches **single-flight**: the first caller starts the build and stores the in-flight promise; concurrent callers `await` the same promise. On settle the in-flight promise is cleared, so a **failed introspection is retryable** instead of caching a rejected promise. `invalidateToolCache()` now also drops any in-flight build.

- `getOrBuildSchema()` → thin single-flight wrapper around a new `buildSchema()`.
- `getSchemaTools()` → thin single-flight wrapper around a new `buildSchemaTools()`.

## Why

Severity is Low (idempotent build, no correctness issue) but it wastes a duplicated PostGraphile introspection round-trip + CPU on tool generation when several admin requests race a cold cache right after boot. Single-flight collapses that to one build. See #367.

## Validation

- `pnpm --filter @vicoop-bridge/server typecheck` — clean.
- New `src/schema-tools.test.ts` (mocks `fetch` with a minimal introspection payload):
  - concurrent first calls trigger **exactly one** introspection and share one cached object;
  - warm cache performs no further fetch;
  - `invalidateToolCache()` forces a fresh introspection;
  - a failed introspection is retryable (in-flight promise cleared on error).
- Confirmed the single-flight test **fails** against the pre-fix code (introspection ran 5×) and passes after.

No changeset: `@vicoop-bridge/server` is in the changeset `ignore` list (ships via `fly deploy`, no published artifact).

Closes #367.

### DIFF
--- a/packages/server/src/schema-tools.test.ts
+++ b/packages/server/src/schema-tools.test.ts
@@ -98,12 +98,13 @@ test('invalidateToolCache forces a fresh introspection', async (t) => {
 
 test('invalidation mid-build does not resurrect the stale result into the cache', async (t) => {
   invalidateToolCache();
-  // Two distinct schemas so we can tell which build populated the cache. The
-  // first build is invalidated while its fetch is still in flight; when it
-  // resolves it must NOT write its (now stale) result back into cachedTools.
-  // Build #1's fetch is deliberately slow (40ms) so it resolves *after* the
-  // post-invalidation build #2 (5ms). Without the epoch guard, build #1's late
-  // .then would overwrite cachedTools with its stale result — the bug.
+  // Both builds use the same payload; we distinguish them by the object
+  // identity of the tool set each build() returns. Build #1 is invalidated
+  // while its fetch is still in flight; when it resolves it must NOT write its
+  // (now stale) result back into cachedTools. Build #1's fetch is deliberately
+  // slow (40ms) so it resolves *after* the post-invalidation build #2 (5ms).
+  // Without the epoch guard, build #1's late .then would overwrite cachedTools
+  // with its stale result — the bug.
   const { fn, callCount } = deferredFetch(
     () => EMPTY_SCHEMA,
     (i) => (i === 0 ? 40 : 5),
--- a/packages/server/src/schema-tools.test.ts
+++ b/packages/server/src/schema-tools.test.ts
@@ -1,0 +1,108 @@
+// Unit tests for the schema-tools cache, focused on the single-flight
+// guarantee from #367: a cold cache hit by concurrent callers must trigger
+// exactly one PostGraphile introspection, and a failed introspection must be
+// retryable rather than poisoning the cache.
+
+import { test } from 'node:test';
+import assert from 'node:assert/strict';
+import { getSchemaTools, invalidateToolCache } from './schema-tools.js';
+
+// Minimal but valid introspection payload: an empty schema is enough to drive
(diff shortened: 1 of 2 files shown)


longfin:
Ran an independent subagent review. Verdict was **ship-with-nits** with one real-but-latent finding, now addressed:

- **[latent bug] mid-build invalidation could resurrect stale data.** If `invalidateToolCache()` fired while a build was parked on its introspection `fetch`, the orphaned build's `.then` would resolve later and write its stale result back into `cachedSchema`/`cachedTools`. `invalidateToolCache` has no production caller today (so this was latent), but it's a footgun for any future schema-reload path. Fixed in 179d86d with a `cacheGeneration` epoch guard: a build only adopts its result / clears its in-flight slot if the generation is still current. Added a regression test (`invalidation mid-build does not resurrect the stale result into the cache`) that slows build #1's fetch so it settles after a post-invalidation build #2 — it fails without the guard.

Other review notes (no code change needed):
- TDZ on `toolsInFlight` referenced in `invalidateToolCache` is safe — the function body only runs after module evaluation.
- Strengthened the single-flight test to assert the concurrent callers share **one in-flight promise** (identity), not just the resolved-value identity, so it can't pass via a warm-cache read.
- Error path leaves the caches untouched and clears the in-flight slot → retryable (covered by test).

Tests: 4/4 pass, typecheck clean.

PR #402, opened 2026-06-26
fix(vicoop-codex): tag context-window overflows with context_length_exceeded

longfin:
## What

When `vicoop-codex serve` relays an upstream **"Your input exceeds the context window of this model"** error as an in-band SSE error frame, the bridge now classifies it from the message and tags the `task.fail` with the standard OpenAI code **`context_length_exceeded`** instead of the generic `upstream_error`. Any other in-band error stays `upstream_error`.

The codec (`@planetarium/oai2a2a-codec`) passes this terminal code through verbatim into the OpenAI error envelope (`mapA2ATerminalError` reads the bridge's `terminal_error.code`), so OpenAI-SDK callers receive `code: "context_length_exceeded"`.

## Why

Follow-up to the in-band-error-frame fix (#400). Standardising the code lets OpenAI-SDK clients recognise a context overflow instead of an opaque upstream failure:
- **Non-streaming opencode** classifies a 502 context-overflow via `parseAPICallError` (matches both the message and `error.code === "context_length_exceeded"`) → can compact-and-retry.
- Any client keying on the OpenAI error code gets a meaningful, standard value.

## Scope note (verified empirically)

This does **not** make *streaming* opencode auto-recover. `@ai-sdk/openai-compatible` collapses an in-band `{error:{...}}` chunk to just its message string (drops `code`/`type`) before opencode sees it, and opencode's `parseStreamError` is code/type-based (never message-matches). So no bridge/router/codec change reaches opencode's classifier on the stream path — that gap is upstream in opencode/the AI SDK. Tracked separately:
- Proactive route (the working lever for opencode): planetarium/vicoop-router#124 (serve `/.well-known/opencode` with per-model `limit.context`), or client-side `opencode.json` `limit`.

This PR is the correct, low-cost standardisation regardless; it's semantically right and helps non-streaming / non-AI-SDK clients.

## Test

- `handle: in-band context-window error frame → task.fail context_length_exceeded, not empty complete`
- `handle: a NON-context in-band error frame stays upstream_error`

Full `vicoop-codex` suite green; `typecheck` clean.

### DIFF
--- a/.changeset/vicoop-codex-context-overflow-code.md
+++ b/.changeset/vicoop-codex-context-overflow-code.md
@@ -1,0 +1,14 @@
+---
+"@vicoop-bridge/client": patch
+---
+
+fix(vicoop-codex): tag context-window overflows with `context_length_exceeded`
+
+When `vicoop-codex serve` relays an upstream "input exceeds the context window"
+error, the bridge now tags the resulting `task.fail` with the standard
+`context_length_exceeded` code instead of the generic `upstream_error`. The
+codec passes this terminal code through verbatim into the OpenAI error
+envelope, so OpenAI-SDK callers can recognise it as a context overflow (e.g.
+non-streaming opencode classifies it via `parseAPICallError` and can
+compact-and-retry) rather than treating it as an opaque upstream failure. Any
+other in-band error stays `upstream_error`.
--- a/packages/client/src/backends/vicoop-codex.test.ts
+++ b/packages/client/src/backends/vicoop-codex.test.ts
@@ -1219,12 +1219,14 @@ test('handle: non-2xx from serve → task.fail upstream_error', async () => {
   assert.ok(failFrame.error.message.includes('overloaded'));
 });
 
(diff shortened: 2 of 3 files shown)


longfin:
Superseded by #398, closing.

#398's rescope moved the overflow matcher into the shared `normalizeTaskFailError` (`failure-code.ts`), and the codex in-band error frame already reaches it as `upstream_error` + the relayed upstream message — so with `exceeds the context window` added to the shared pattern list, the exact behavior of this PR (in-band context-window frame → `task.fail` with `context_length_exceeded`, non-context frames stay `upstream_error`) falls out of the shared path with no codex-backend change. Both backend-level tests from this PR are ported into #398 verbatim (`vicoop-codex.test.ts`), and `context_length_exceeded` is preserved through normalization there as well.

One shared classifier instead of per-backend matchers also keeps the code list in one place, consistent with the gateway-side mapping in planetarium/oai2a2a#115.

PR #416, opened 2026-07-08
fix(server): guarantee terminal frames close the SSE stream (identity-safe bindings + inactivity backstop)

longfin:
## Problem

An OpenAI-compatible client (via the router) intermittently observed the **full response arrive but the connection never close**. Root cause is on the bridge: a task's SSE stream only closes when its executor observes a terminal event, but the terminal frame could be **silently lost** — after which the executor blocks forever on `AsyncEventQueue.iterate()` with **no timeout, no log, and no rescue but a process restart**.

Chain: OpenAI client → router (`/v1` streaming: oai2a2a codec + `withKeepAlive`) → **bridge (A2A SSE)** → daemon (WS). The daemon behaves correctly; the terminal is lost downstream of it.

## Two ways the terminal went missing

1. **Binding clobber.** `registry.unbindTask(taskId)` deleted by taskId with **no identity check**, and the executor called it in a `finally` that runs **after** an awaited `taskStore.updateTask` transaction. On a taskId reused across turns (the `input-required` continuation, kept on the same taskId per the A2A spec), a stale prior-turn executor's late unbind deleted the **next** turn's freshly-installed binding. `ws.ts`'s `task.complete`/`task.fail` handlers then hit `getBinding() === undefined` and returned silently, dropping the terminal → executor wedged forever.
2. **Live-binding overwrite.** `bindTask` blindly overwrote a live binding for an in-flight task on a duplicate/retried request, orphaning the first executor's queue (deterministic, no timing window).

Neither had any observability, timeout, or rescue (`tasks/cancel` and agent disconnect both fail to reach a clobbered binding).

## Fixes

- **Identity-scoped `unbindTask(taskId, binding)`** — only deletes when the entry is still the caller's binding. Mirrors the ws-identity guard already applied to the agents map (#365). Same guard applied to `abortControllers`.
- **`bindTask` terminates a displaced live binding** with a `failed` (`task_superseded`) terminal so its stream closes instead of orphaning; emits `binding_displaced`.
- **Server-authoritative inactivity backstop** in `WSForwardingExecutor`: if a bound task produces no frame (content / artifact / heartbeat) for `BRIDGE_TASK_INACTIVITY_TIMEOUT_MS` (default 10 min, `0` disables), fabricate a `failed` terminal so the stream closes. **Every inbound frame resets it**, so a healthily-heartbeating long task is never reaped. Covers any future terminal-loss path, plus the currently-live case of a backend that resolves without a terminal frame while keeping the socket open.
- **Observability:** `dropped_terminal_frame` on the previously-silent missing-binding return in the ws.ts terminal handlers.

## Tests

New: identity-unbind non-clobber, self-rebind no-op, `bindTask` displacement emits superseded terminal, watchdog closes a silent stream with a `failed` terminal and releases the binding.

Server suite: **232 pass / 0 fail** (`pnpm --filter @vicoop-bridge/server test`), typecheck clean.

## Notes / out of scope

- New env `BRIDGE_TASK_INACTIVITY_TIMEOUT_MS` documented in `fly.toml`; three new log events added to the `vicoop-bridge-logs` skill table.
- **Router-side complement (separate repo):** the router's stall watchdog is reset by A2A heartbeat bytes (`sse.ts`, #95-related), so heartbeats can defeat the only OAI-side backstop. This bridge fix removes the terminal-loss at the source; a router-side hard cap decoupled from heartbeats is recommended as defense-in-depth.
- Not included (watchdog covers them, but not root-cleaned): treating a non-terminal `input-required` sent via `task.status` as terminal, and the `input-required` turn persisting a stale pre-turn status to the task store.


longfin:
Added a second commit with production instrumentation for the *"response arrived but the stream never closed"* class (the actual user report, via opencode). Investigation notes:

- Repro attempts (router `pnpm e2e --diag --stream` against `claude-sonnet-5`: input-required ×8, multi-turn tool loop ×6) all closed **cleanly** (`finish_reason` + `[DONE]`) — the common paths are fine, so the hang is a rare tail.
- Provider daemon logs (`vicoop-claude-baseline`) show taskIds are **unique per turn** (no reuse) and the daemon terminates **3796/3799** tasks — so the binding-race and no-terminal-daemon paths aren't the driver in this deployment. The tail cause is not yet reproduced.

So this PR now leans on **detection + a bounded backstop** rather than claiming a single root cause. Three events make a bridge-side hang observable/bounded once deployed (`fly deploy`):

- `task_inactivity_timeout` — watchdog reaped a silent stream (the direct "would-have-hung" signal; also bounds any hang to `BRIDGE_TASK_INACTIVITY_TIMEOUT_MS`, default 10m).
- `dropped_terminal_frame` — a terminal arrived for a taskId with no live binding.
- `task_stream_closed` (new) — a stream closed **without** ever receiving a client-marked-`final` frame (`reason=aborted|no_final_frame`, `durationMs`, `finalState`). Low-noise: clean terminals don't emit it.

Separately surfaced (not fixed here): the prod `outcome=timeout` rows are a *different* issue — long non-streaming claude tasks (~6 min, `elapsedMs` 360–390k) exceeding the router's 300s stall budget; the client gets a stall error, not a hang. Fix for that belongs on the claude backend (heartbeat/stream during long runs) and/or the router stall budget.

PR #421, opened 2026-07-13
feat(client): advertise per-model contextWindow / maxOutputTokens on openai-compat/v1

longfin:
Implements the producer side of the openai-compat/v1 `contextWindow` / `maxOutputTokens` advertise (spec: planetarium/oai2a2a#117).

## What

Each backend now populates the two token-limit hints on its `params.models[]` advertise from its **own** model metadata — detected, not hardcoded:

| Backend | contextWindow source | maxOutputTokens source |
|---|---|---|
| **codex** | `model/list` `context_window` (effective window) | — (codex exposes no output ceiling) |
| **claude** | Anthropic `GET /v1/models` `max_input_tokens` | `max_tokens` |

**claude auth**: the Models API is fetched with the host **subscription OAuth token** — the same cred + `oauth-2025-04-20` header the `usage()` path already reads. Verified live: the token returns `200` (no separate `ANTHROPIC_API_KEY` needed).

**Tier correction (option B)**: `max_input_tokens` is the model's *ceiling* (1M for a 1M-capable model), but the effective window is only that large when the 1M tier is active. So an advertised id is capped at the 200k base unless it carries the `[1m]` marker. `max_tokens` is tier-agnostic.

## Design

- Enrichment is **opt-in** via a `resolveModelLimits` seam (wired in `cli.ts` to `loadClaudeModelCatalog`) — existing exact-match `resolveCapabilities` tests keep asserting the bare id/reasoning/default advertise, and nothing reaches the network or host keychain implicitly.
- **Best-effort**: a missing token, failed lookup, or unknown id simply omits the hints (spec-valid — absence means "unspecified").
- The protocol `OpenAICompatModelAdvertise` schema is a strict `z.object`, so the two fields are declared there explicitly (otherwise they'd be stripped before the wire).

## Tests

+18 (protocol 2, codex 1, claude enrich/resolve 8, claude-models fetch 7). Full suite green: client 820, server 233, protocol 4; typecheck clean.

## Notes

- `vicoop-codex` / `openclaw` backends still omit the hints (spec-valid) — follow-up if desired.
- Consumer side (a2x-internal-router: parse → store → expose on `/v1/models`) is a separate PR.


longfin:
Follow-up commit (74a123d) adds the **vicoop-codex** backend to this PR — the earlier "vicoop-codex omits the hints (follow-up)" note is now resolved.

Investigation showed vicoop-codex exposed no window through any interface (`models --json` returned only `id` + `service_tiers`). But the underlying ChatGPT Codex backend `/models` response **does** carry `context_window` — the CLI's `normalizeModel` was just dropping it. Fixed upstream in **planetarium/vicoop-codex-cli#34**; this commit parses that field in `probeVicoopCodexModels` (reshaped `string[]` → `{ id, contextWindow }[]`) and advertises `contextWindow`.

Guarded + null-when-absent, so it's a no-op against a vicoop-codex predating the CLI change and activates once that release lands. No output-token ceiling exists for this backend, so `maxOutputTokens` stays unadvertised (same as the `codex` backend).

**Merge/deploy order**: vicoop-codex-cli#34 releases first → deployed vicoop-codex binary surfaces `context_window` → this backend's advertise populates. Until then it's inert (as designed).

PR #427, opened 2026-07-15
fix(client): reap vicoop-codex serve child on every non-SIGKILL exit

longfin:
## Why

Investigating a process leak on an operator host: **4 `vicoop-codex serve` processes had been running for 14 days, orphaned (reparented to init, ppid=1), still LISTENing on ephemeral ports, with no live `vicoop-client` parent** — one accumulated per client restart.

The `vicoop-codex` host backend spawns a long-lived `vicoop-codex serve` app-server (a lazy singleton). Its **only** teardown is `backend.stop()` (`backends/vicoop-codex.ts`), reached via `Client.stop()`. But `client.stop()` is only called from the SIGINT/SIGTERM handler (`onSignal`). The daemon has **several other exit paths that bypass `onSignal`**:

- **Fatal terminal WS close** → `onFatal: () => process.exit(1)` (e.g. 4014 "client deleted") — a normal operational event
- **`uncaughtException` / `unhandledRejection`** → `process.exit(1)`
- **SIGHUP** — not registered (only SIGINT/SIGTERM were), so Node's default disposition terminates the daemon
- **SIGKILL** — untrappable

The serve child is spawned **without `detached`**, and `process.exit()` sends no signal to children, so on all of the above the child was orphaned and kept LISTENing forever.

## What

- Register **SIGHUP** through the same graceful `onSignal` path (covers a torn-down controlling terminal / login session, e.g. under a launcher like cmux).
- Add a **`process.once('exit')` last-resort** that calls `client.stop()` synchronously. `process.exit()` runs `'exit'` listeners synchronously, and `client.stop()` → `backend.stop()` → `child.kill('SIGTERM')` is synchronous and idempotent, so the child is **reliably reaped on every non-SIGKILL exit** (`onFatal`, uncaught/unhandled, normal exit).

SIGKILL can't be trapped — the complementary **parent-death watchdog** in `vicoop-codex serve` (vicoop-codex-cli) covers that case (and hard crashes) by self-terminating when its parent disappears.

## Test

`packages/client` typecheck clean; `client.test.ts` + `cli-args.test.ts` 80/80 pass. The change is confined to `runDaemon`'s signal/exit wiring in `cli.ts`.

### DIFF
--- a/.changeset/vicoop-codex-serve-leak-on-exit.md
+++ b/.changeset/vicoop-codex-serve-leak-on-exit.md
@@ -1,0 +1,19 @@
+---
+"@vicoop-bridge/client": patch
+---
+
+client daemon: reap the vicoop-codex `serve` child on every non-SIGKILL exit
+
+The `vicoop-codex` host backend spawns a long-lived `vicoop-codex serve`
+app-server whose only teardown is `backend.stop()`, reached via
+`client.stop()`. But the daemon also exits WITHOUT running the SIGINT/SIGTERM
+handler — on a fatal terminal WS close (`onFatal` → `process.exit(1)`), an
+`uncaughtException`/`unhandledRejection`, or SIGHUP (previously unhandled) —
+so `stop()` never fired and the serve child leaked as an orphan (reparented to
+init, still LISTENing), accumulating one per restart.
+
+Register SIGHUP through the graceful path, and add a `process.once('exit')`
+last-resort that calls `client.stop()` synchronously — it runs on every
+`process.exit()` path, so the child is reliably reaped on all non-SIGKILL
+exits. (SIGKILL can't be trapped; vicoop-codex-cli's `serve` carries its own
+parent-death watchdog for that case.)
--- a/packages/client/src/cli.ts
+++ b/packages/client/src/cli.ts
@@ -697,6 +697,37 @@ async function runDaemon(parsed: Extract<CliArgs, { action: 'daemon' }>): Promis
   };
   process.on('SIGINT', onSignal);
   process.on('SIGTERM', onSignal);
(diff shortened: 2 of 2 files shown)


longfin:
### Adversarial safety review (3 independent passes) — any unwanted teardown?

**Verdict: safe, no regression.** Findings:

- **Double-stop is idempotent.** On SIGINT/SIGTERM, `onSignal` calls `client.stop()` then `process.exit(0)`, which now also fires the `'exit'` handler → `client.stop()` again. Traced: `stopped` guard, `AbortController.abort()` / `ws.close()` are no-ops on repeat, and `backend.stop()` sets `serve = null` so the second kill is a clean no-op.
- **SIGHUP is not a "was-surviving-now-dies" regression.** Verified on Node v24.4.1: Node resets an inherited `SIG_IGN`, so a delivered SIGHUP **already** terminated the daemon (abruptly, orphaning the child). This change only makes that same death *graceful* (reaps the child). The recommended `start --detach` path is a session leader with no controlling terminal, so no SIGHUP reaches it. No SIGHUP-as-reload semantics exist in the codebase.
- **Scope:** `backend.stop()` signals only *this* daemon's own `serve` child pid (not a process group) — nothing shared/external. `runDaemon` is the sole caller (no tests/embedders), and `process.once` means no listener accumulation.

**Hardening applied from the review:** the reviewer noted `client.stop()` reaches `backend.stop()` *last* (after aborting inflight + `ws.close()`), so a synchronous throw in an earlier step could skip the reap inside the `try/catch`. The `'exit'` handler now calls `backend.stop?.()` **first and independently**, then `client.stop()` — guaranteeing the child is reaped regardless of `client.stop()` internals.

Complements planetarium/vicoop-codex-cli#43 (SIGKILL / hard-crash coverage via serve's own parent-death watchdog).

PR #450, opened 2026-08-05
feat(server): charge per call with x402 (exact + metered upto)

longfin:
Rebased onto `main` now that #449 (the `@a2x/sdk` 0.18 bump this needs) has merged. Two commits: `exact` first, then `upto` on top — reviewable separately, though the second reshapes the pricing column the first introduced.

Both schemes ship together on purpose: the `agents.x402_pricing` migration lands once, with its final shape, instead of being reshaped a release later.

## What this adds

A per-agent x402 payment gate in front of the WS-forwarding path. An agent with no pricing row is free and takes no payment code path at all — `currentGate()` returns `undefined` and nothing else in the executor changes.

```
turn 1  unpaid request      → input-required + x402.payment.required
turn 2  signed resubmission → verify → forward to agent → settle
```

| scheme | caller pays | notes |
| --- | --- | --- |
| `exact` (default) | flat fee per call | works on every backend |
| `upto` | tokens actually consumed, up to an authorized ceiling | x402 V2 only; needs caller opt-in; not possible on openclaw |

Two properties fall out of the ordering, and they're the reason for it:

- **Unpaid calls never reach the agent.** The gate runs before the task is bound or forwarded, so an unpaid or malformed request costs the backend nothing.
- **Failed work is never charged.** Settlement runs only on a `completed` task; a failed, cancelled, or timed-out task leaves the payer's signed authorization unused. A settlement that itself fails turns the task `failed` rather than letting it report a success the merchant wasn't paid for. (Under `upto`, deferring settlement is also what makes metering possible — the token counts only exist once the work is done.)

## Metering needed no protocol change

Worth flagging because I previously claimed the opposite in this PR's own docs, and it was wrong.

The claude, codex, and vicoop-codex backends **already** stamp per-turn token counts onto `task.complete`, under the openai-compat extension URI in `status.message.metadata` — deliberately even when the caller didn't activate that extension, so that "billing telemetry" can read them (`openai-compat-usage.ts:57-63`). `wireMessageToA2X` (`ws.ts:193`) carries the metadata through untouched. The counts are therefore already on the terminal event `settleTerminal` receives; `readTaskUsage` just reads them, from either wire shape (`{usage}` or `{chat_completion:{usage}}`).

No backend change, no protocol change, no `TaskCompleteFrame` field.

This is **not** the `usage.request` RPC — that's a cumulative account-level quota snapshot (rolling subscription windows, rate-limited, 5-min cached) and is useless for metering one call.

## Design decisions worth reviewing

**What to charge when a completed call can't be priced.** The load-bearing judgement call. openclaw reports no usage at any layer; codex and vicoop-codex emit `{0,0,0}` when their runtime drops accounting for a turn. **Zero does not mean free.** The bridge charges the configured `minAmount` (zero if unset) and logs `x402_usage_unavailable` with what it charged.

The default direction is deliberately payer-favourable: billing the authorized ceiling because *our* instrumentation lost the token count would be a user-visible wrong, whereas undercharging is our own loss to fix. It's one config field, so the opposite policy is a one-line change if you'd rather. The docs tell operators plainly to set `minAmount` on metered agents.

**Pricing is DB-owned, not declared by the client.** `payTo` names the wallet that receives money and the `hello` frame is authored by the connecting agent, so pricing gets the same trust boundary as `allowed_callers`. A malformed row logs `x402_pricing_invalid` and connects the agent as **free** rather than taking it offline — an operator typo shouldn't become an outage, and nobody is charged against config the server couldn't parse.

**Offerings persist in Postgres, not the SDK's `InMemoryX402Store`.** Correctness, not optimization: the round-trip spans two HTTP requests and the bridge runs several Fly instances behind a load balancer, so turn 2 routinely lands on an instance that never saw turn 1. In memory that classifies as `no-stored-offering` and refuses a perfectly valid payment. Expiry is lazy (a `WHERE` clause), per the store contract — no reaper needed.

**Rates are per million tokens, in atomic units, BigInt throughout.** That's the unit vendors quote in, so a price list transcribes directly. The intermediate product leaves a double's exact range long before the amounts do, hence BigInt end to end. `cached_tokens` is a breakdown of `prompt_tokens` and `reasoning_tokens` of `completion_tokens` — neither additive, so neither double-counted. An absent `cachedInput` rate means "same as input", not "free"; treating an unstated discount as 100% would give away most of a cache-heavy call.

**Driving `X402Context` from a custom executor.** The SDK documents this flow inside `BaseAgent.run()`, which the bridge has no room for — `WSForwardingExecutor` overrides `execute`/`executeStream` outright. `X402Context`'s methods take a structural `{ taskId, message }` rather than a real `InvocationContext`, so they compose unchanged; only the `AgentEvent`s need translating into wire status events.

**`X402Gate` takes an injected context**, with `createPostgresX402Context()` building the production one. That's what makes the round-trip testable against a stub facilitator instead of a live chain.

## Two prerequisites for `upto` (not blockers for merging)

- **x402 V2.** We default to V2. If `BRIDGE_X402_VERSION=1`, an `upto` agent refuses every call with `invalid_x402_version` and logs `x402_config_invalid` — reported as the permanent misconfiguration it is, rather than as a transient outage.
- **Caller opt-in.** Signing `upto` authorizes spending up to the maximum at the merchant's discretion — broader consent than `exact` — so the SDK's default selector never auto-picks one. Clients must pass `allowUpto` or select the requirement explicitly. **An `upto`-only agent is unusable by a default-configured client**, so we should confirm `a2a-wallet` exposes it before switching any existing agent over.

## Also

Drops the dead `x402@1.2.0` dependency. SDK 0.16 replaced it with `@x402/core` + `@x402/evm`, and nothing in the repo imported it — it was also the source of the `wagmi`→`react` peer warning on install.

## Test plan

- [x] `pnpm -r build`, `pnpm -r typecheck`
- [x] `pnpm --filter @vicoop-bridge/server test` — 335 tests, 280 pass / 0 fail (55 skipped, env-gated)
- [x] `pnpm --filter @vicoop-bridge/client test` — 858 pass / 0 fail

New coverage (34 tests):

- `x402/gate.test.ts` (16) — the full round-trip against a stub facilitator that counts `verify`/`settle` calls and records the settled amount, so "not charged for failed work" and "settled the metered amount, not the ceiling" are asserted directly rather than implied. Includes every refusal path: unknown taskId, a payload redirecting funds to a different `payTo` (caught locally, before facilitator quota is spent), facilitator rejection, settle failure, settle throwing, and a payment-rail outage failing **closed**. On the metered side: a charge above the ceiling is confirmed clamped down to it, and an `exact` agent is confirmed to ignore usage entirely (metering an `exact` requirement throws in the SDK).
- `x402/pricing.test.ts` (19) — schema and rate arithmetic. `"0.01"` (dollars-not-atomic-units) and `10000` (a number, precision-lossy on 18-decimal assets) are both rejected; cache discounting doesn't double-count; a sub-unit charge rounds up rather than to zero; BigInt exactness holds on values past 2^53.
- `x402/usage.test.ts` (7) — both wire shapes, `total_tokens` recomputed rather than trusted, an incoherent cache count dropped instead of discounted on, and `{0,0,0}` surfaced as a real zero rather than as absent.
- `executor.test.ts` (+1) — a free agent is forwarded normally even when the payment path is configured, with an exploding `sql` proxy proving the store is never touched.

Not covered by automated tests: `PostgresX402Store` against a real database, and real on-chain settlement (both `exact` EIP-3009 and `upto` Permit2). Those want the Base Sepolia sandbox — see `docs/x402-payments.md`.

### DIFF
--- a/packages/server/src/admin-api.ts
+++ b/packages/server/src/admin-api.ts
@@ -9,7 +9,12 @@ import type { Registry } from './registry.js';
 import { validatePrincipal } from './auth/principal.js';
 import { generateApiKeyId, issueSessionToken } from './auth/caller-token.js';
 import { isAdmin } from './admin-scope.js';
-import { X402PricingSchema, parseX402Pricing, type X402Pricing } from './x402/pricing.js';
+import {
+  X402PricingWriteSchema,
+  formatPricingError,
+  parseX402Pricing,
+  type X402Pricing,
+} from './x402/pricing.js';
 
 type Tx = postgres.TransactionSql;
 
@@ -268,11 +273,14 @@ export async function setX402Pricing(
   agentId: string,
   pricing: unknown,
 ): Promise<X402PricingResult> {
+  // The strict schema: an unrecognized key fails here rather than being
+  // dropped. Every optional field on a pricing object changes what is
+  // charged, so a silently-ignored typo is a silently-wrong price.
   let parsed: X402Pricing;
   try {
-    parsed = X402PricingSchema.parse(pricing);
+    parsed = X402PricingWriteSchema.parse(pricing);
   } catch (err) {
-    throw new AdminApiError(`Invalid x402 pricing: ${String(err)}`, 400);
+    throw new AdminApiError(`Invalid x402 pricing: ${formatPricingError(err)}`, 400);
   }
 
   const rows = await db.begin(async (tx) => {
--- a/packages/server/src/admin-api.ts
+++ b/packages/server/src/admin-api.ts
@@ -15,6 +15,7 @@ import {
   parseX402Pricing,
   type X402Pricing,
 } from './x402/pricing.js';
+import { notifyX402PricingChanged } from './x402/pricing-watch.js';
 
 type Tx = postgres.TransactionSql;
 
@@ -298,7 +299,11 @@ export async function setX402Pricing(
   // distinguishing them would confirm the existence of someone else's agent.
   if (rows.length === 0) throw new AdminApiError('Agent not found or not authorized.', 404);
 
+  // Patch this instance, then tell the others. The agent's WebSocket usually
+  // lives on a different instance than the one that served this request, and
+  // that one is the one whose cached pricing decides what the agent charges.
   registry.updateX402Pricing(agentId, parsed);
+  await notifyX402PricingChanged(db, agentId);
   return { agent_id: agentId, x402_pricing: parsed };
 }
 
@@ -319,7 +324,10 @@ export async function clearX402Pricing(
   });
   if (rows.length === 0) throw new AdminApiError('Agent not found or not authorized.', 404);
 
+  // The direction that matters most: an agent made free here must stop being
+  // billed everywhere, not just on this instance.
   registry.updateX402Pricing(agentId, undefined);
+  await notifyX402PricingChanged(db, agentId);
   return { agent_id: agentId, x402_pricing: null };
 }
 
--- a/packages/server/src/admin-api.ts
+++ b/packages/server/src/admin-api.ts
@@ -9,6 +9,7 @@ import type { Registry } from './registry.js';
 import { validatePrincipal } from './auth/principal.js';
 import { generateApiKeyId, issueSessionToken } from './auth/caller-token.js';
 import { isAdmin } from './admin-scope.js';
+import { X402PricingSchema, parseX402Pricing, type X402Pricing } from './x402/pricing.js';
 
 type Tx = postgres.TransactionSql;
 
@@ -217,6 +218,101 @@ export async function listCallers(
     created_at: policy.created_at instanceof Date ? policy.created_at.toISOString() : policy.created_at as string | undefined,
     updated_at: policy.updated_at instanceof Date ? policy.updated_at.toISOString() : policy.updated_at as string | undefined,
   };
+}
+
+// ----- x402 pricing ---------------------------------------------------------
+//
+// `agents.x402_pricing` is DB-owned rather than declared in the WS hello
+// frame, because `payTo` names the wallet that receives money and the hello
+// frame is authored by the connecting agent. That makes this the only write
+// path, so it validates through the same `parseX402Pricing` the runtime uses:
+// a bad address or a dollars-instead-of-atomic-units amount is rejected here,
+// at write time, rather than silently disabling payments at the agent's next
+// connect.
+
+export interface X402PricingResult {
+  agent_id: string;
+  x402_pricing: X402Pricing | null;
+}
+
+export async function getX402Pricing(
+  db: Sql,
+  principalId: string,
+  agentId: string,
+): Promise<X402PricingResult> {
+  const rows = await db.begin(async (tx) => {
+    await setRlsContext(tx, principalId);
+    return tx`SELECT id AS agent_id, x402_pricing FROM agents WHERE id = ${agentId}`;
+  });
+  if (rows.length === 0) throw new AdminApiError('Agent not found.', 404);
+
+  // A row that fails to parse is reported rather than thrown: the operator
+  // asking "what does this agent charge?" is exactly who needs to be told it
+  // is currently unreadable (and therefore being served for free).
+  let pricing: X402Pricing | null = null;
+  try {
+    pricing = parseX402Pricing(rows[0].x402_pricing) ?? null;
+  } catch (err) {
+    throw new AdminApiError(
+      `Stored pricing for this agent is invalid and is being ignored at runtime: ${String(err)}`,
+      409,
+    );
+  }
+  return { agent_id: agentId, x402_pricing: pricing };
+}
+
+export async function setX402Pricing(
+  db: Sql,
+  registry: Registry,
+  principalId: string,
+  agentId: string,
+  pricing: unknown,
+): Promise<X402PricingResult> {
+  let parsed: X402Pricing;
+  try {
+    parsed = X402PricingSchema.parse(pricing);
+  } catch (err) {
+    throw new AdminApiError(`Invalid x402 pricing: ${String(err)}`, 400);
+  }
+
+  const rows = await db.begin(async (tx) => {
+    await setRlsContext(tx, principalId);
+    return tx`
+      UPDATE agents
+      SET x402_pricing = ${tx.json(JSON.parse(JSON.stringify(parsed)))},
+          updated_at = now()
+      WHERE id = ${agentId}
+      RETURNING id AS agent_id
+    `;
+  });
+  // RLS hides agents the operator doesn't own, so zero rows is
+  // indistinguishable from "no such agent" — and deliberately so, since
+  // distinguishing them would confirm the existence of someone else's agent.
+  if (rows.length === 0) throw new AdminApiError('Agent not found or not authorized.', 404);
+
+  registry.updateX402Pricing(agentId, parsed);
+  return { agent_id: agentId, x402_pricing: parsed };
+}
+
+export async function clearX402Pricing(
+  db: Sql,
+  registry: Registry,
+  principalId: string,
+  agentId: string,
+): Promise<X402PricingResult> {
+  const rows = await db.begin(async (tx) => {
+    await setRlsContext(tx, principalId);
+    return tx`
+      UPDATE agents
+      SET x402_pricing = NULL, updated_at = now()
+      WHERE id = ${agentId}
+      RETURNING id AS agent_id
+    `;
+  });
+  if (rows.length === 0) throw new AdminApiError('Agent not found or not authorized.', 404);
+
+  registry.updateX402Pricing(agentId, undefined);
+  return { agent_id: agentId, x402_pricing: null };
 }
 
 // ----- Static API keys (issue #308) -----------------------------------------
--- a/packages/server/src/agent-card.ts
+++ b/packages/server/src/agent-card.ts
@@ -7,7 +7,9 @@ import {
 import { SIWE_BEARER_AUTH_EXTENSION_URI } from '@vicoop-bridge/protocol';
 import type { ClientConnection, Registry } from './registry.js';
 import { WSForwardingExecutor } from './executor.js';
+import { isX402ExtensionUri } from '@a2x/sdk/x402';
 import { X402_FOUNDATION_EXTENSION_URI } from './x402/gate.js';
+import { logEvent } from './log.js';
 import type { Sql } from './db.js';
 
 export interface AgentA2XOptions {
@@ -41,12 +43,26 @@ export function buildAgentA2XServer(
     ? `${opts.publicUrl}/agents/${conn.agentId}`
     : `/agents/${conn.agentId}`;
 
+  // x402 `resource` must be an absolute URL — strict facilitators reject
+  // anything else, so without `publicUrl` a paid agent would publish offerings
+  // that can never verify or settle. Withhold the gate instead: the agent
+  // serves for free and says so in the log, which is a working deployment with
+  // a visible misconfiguration rather than a broken payment loop. Same posture
+  // as an unparseable pricing row.
+  const paymentsWired = Boolean(opts.db) && Boolean(opts.publicUrl);
+  if (opts.db && !opts.publicUrl && conn.x402Pricing) {
+    logEvent('x402_config_invalid', {
+      agentId: conn.agentId,
+      reason: 'pricing is configured but PUBLIC_URL is not, so no absolute resource URL exists',
+    });
+  }
+
   const executor = new WSForwardingExecutor(
     conn.agentId,
     registry,
     taskStore,
     undefined,
-    opts.db ? { sql: opts.db, resource: url } : undefined,
+    paymentsWired ? { sql: opts.db!, resource: url } : undefined,
   );
 
   const a2xServer = new A2XServer({
@@ -105,6 +121,18 @@ export function buildAgentA2XServer(
       // publicUrl).
       continue;
     }
+    if (isX402ExtensionUri(extension.uri)) {
+      // Always dropped, paid or not. `addExtension` is append-only, so leaving
+      // a wire-declared x402 entry in would let a free agent advertise a price
+      // the bridge will never ask for, and would give a paid agent two entries
+      // for the same URI at different prices — with no rule saying which one a
+      // client reads. Pricing is DB-owned; so is saying so on the card.
+      logEvent('x402_wire_advertisement_dropped', {
+        agentId: conn.agentId,
+        uri: extension.uri,
+      });
+      continue;
+    }
     a2xServer.addExtension(extension);
   }
   if (bridgeWillEmitSiwe) {
@@ -139,7 +167,9 @@ export function buildAgentA2XServer(
   // `input-required` response naming the price, which is more useful than
   // being refused at extension activation. `params` mirrors the offering so a
   // client can decide whether it is willing to pay before spending a turn.
-  if (conn.x402Pricing) {
+  // Gated on `paymentsWired`, not just on pricing being set: the card must not
+  // quote a price the gate is not installed to collect.
+  if (conn.x402Pricing && paymentsWired) {
     const pricing = conn.x402Pricing;
     a2xServer.addExtension({
       uri: X402_FOUNDATION_EXTENSION_URI,
--- a/packages/server/src/agent-card.ts
+++ b/packages/server/src/agent-card.ts
@@ -7,10 +7,15 @@ import {
 import { SIWE_BEARER_AUTH_EXTENSION_URI } from '@vicoop-bridge/protocol';
 import type { ClientConnection, Registry } from './registry.js';
 import { WSForwardingExecutor } from './executor.js';
+import { X402_FOUNDATION_EXTENSION_URI } from './x402/gate.js';
+import type { Sql } from './db.js';
 
 export interface AgentA2XOptions {
   publicUrl: string | undefined;
   deviceFlowEnabled: boolean;
+  // DB handle for the x402 offering store. Absent on deployments assembled
+  // without a payment path; the executor then never installs the gate.
+  db?: Sql;
 }
 
 /**
@@ -36,7 +41,13 @@ export function buildAgentA2XServer(
     ? `${opts.publicUrl}/agents/${conn.agentId}`
     : `/agents/${conn.agentId}`;
 
-  const executor = new WSForwardingExecutor(conn.agentId, registry, taskStore);
+  const executor = new WSForwardingExecutor(
+    conn.agentId,
+    registry,
+    taskStore,
+    undefined,
+    opts.db ? { sql: opts.db, resource: url } : undefined,
+  );
 
   const a2xServer = new A2XServer({
     taskStore,
@@ -115,6 +126,32 @@ export function buildAgentA2XServer(
           mintToken: `a2a-wallet siwe auth --domain ${siweDomain} --uri ${opts.publicUrl} --ttl 1h --json | jq -r '.token'`,
           sendMessage: `a2a-wallet a2a send --bearer "$TOKEN" ${opts.publicUrl}/agents/${conn.agentId}/.well-known/agent-card.json "Hello"`,
         },
+      },
+    });
+  }
+
+  // Advertise x402 when this agent charges. The bridge owns this
+  // advertisement for the same reason it owns the SIWE one: it is the only
+  // side that knows whether a payment gate is actually installed, and pricing
+  // is DB-owned rather than declared by the connecting client.
+  //
+  // `required: false` — a caller that cannot pay still gets a well-formed
+  // `input-required` response naming the price, which is more useful than
+  // being refused at extension activation. `params` mirrors the offering so a
+  // client can decide whether it is willing to pay before spending a turn.
+  if (conn.x402Pricing) {
+    const pricing = conn.x402Pricing;
+    a2xServer.addExtension({
+      uri: X402_FOUNDATION_EXTENSION_URI,
+      description:
+        'x402 payments. Calls are answered with an input-required task carrying x402.payment.required; sign the payload and resubmit against the same taskId.',
+      required: false,
+      params: {
+        network: pricing.network,
+        amount: pricing.amount,
+        asset: pricing.asset,
+        payTo: pricing.payTo,
+        scheme: 'exact',
       },
     });
   }
--- a/packages/server/src/agent-card.ts
+++ b/packages/server/src/agent-card.ts
@@ -144,15 +144,31 @@ export function buildAgentA2XServer(
     a2xServer.addExtension({
       uri: X402_FOUNDATION_EXTENSION_URI,
       description:
-        'x402 payments. Calls are answered with an input-required task carrying x402.payment.required; sign the payload and resubmit against the same taskId.',
+        pricing.scheme === 'upto'
+          ? 'x402 metered payments. Calls are answered with an input-required task carrying x402.payment.required; sign the payload (upto requires opting in on the client) and resubmit against the same taskId. You are charged for the tokens actually consumed, up to the authorized maximum.'
+          : 'x402 payments. Calls are answered with an input-required task carrying x402.payment.required; sign the payload and resubmit against the same taskId.',
       required: false,
-      params: {
-        network: pricing.network,
-        amount: pricing.amount,
-        asset: pricing.asset,
-        payTo: pricing.payTo,
-        scheme: 'exact',
-      },
+      params:
+        pricing.scheme === 'upto'
+          ? {
+              scheme: 'upto',
+              network: pricing.network,
+              asset: pricing.asset,
+              payTo: pricing.payTo,
+              // The ceiling, not the charge — named to match, because a
+              // client that read it as the price would refuse offers it can
+              // comfortably afford.
+              maxAmount: pricing.maxAmount,
+              ratesPerMTok: pricing.rates,
+              ...(pricing.minAmount !== undefined ? { minAmount: pricing.minAmount } : {}),
+            }
+          : {
+              scheme: 'exact',
+              network: pricing.network,
+              amount: pricing.amount,
+              asset: pricing.asset,
+              payTo: pricing.payTo,
+            },
     });
   }
 
--- a/packages/server/src/executor.ts
+++ b/packages/server/src/executor.ts
@@ -19,7 +19,7 @@ import { AsyncEventQueue } from './event-queue.js';
 import { logEvent } from './log.js';
 import { terminalErrorMessageFields } from './terminal-error.js';
 import { X402Gate, createPostgresX402Context } from './x402/gate.js';
-import { readTaskUsage } from './x402/usage.js';
+import { readTaskUsage, type TaskUsage } from './x402/usage.js';
 import type { Sql } from './db.js';
 
 /**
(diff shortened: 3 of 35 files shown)


packages/server/src/x402/store.ts
```diff
@@ -0,0 +1,144 @@
...
+            expires_at = EXCLUDED.expires_at,
+            updated_at = now()
+    `;
+  }
+
+  async get(taskId: string): Promise<X402StoreEntry | undefined> {
+    const rows = await this.sql<OfferingRow[]>`
+      SELECT task_id, entry
+      FROM x402_offerings
+      WHERE task_id = ${taskId}
```
longfin: [P1] Offering이 agent에 scope되지 않습니다. `agent_id`를 row에 기록하지만 `get`/`update`/`delete`는 `task_id`만 사용하고, PK와 `ON CONFLICT`도 `task_id` 단독입니다. 그래서 A의 저렴한 offering과 서명을 같은 `taskId`로 B endpoint에 제출하면 B gate가 A requirement를 valid로 분류해 B 작업을 전달합니다. 로컬 재현에서 B가 999999로 설정되어 있어도 A의 10000 requirement가 승인/settle됐습니다. `(agent_id, task_id)` 복합 키를 사용하고 모든 store 연산을 `this.agentId`로 제한하며, continuation task도 endpoint agent와 일치하는지 확인해야 합니다.

longfin (commented):
결제 처리에 merge 전에 수정해야 할 P1 3건과 운영 이슈 3건을 확인했습니다. 상세 내용은 inline comment에 남깁니다. (PR 작성자 계정이라 REQUEST_CHANGES 상태는 지정할 수 없어 COMMENT review로 제출합니다.)

packages/server/src/x402/gate.ts
```diff
@@ -0,0 +1,437 @@
...
+    | { kind: 'settled'; metadata: Record<string, unknown> }
+    | { kind: 'failed'; event: TaskStatusUpdateEvent }
+  > {
+    const { taskId, contextId, classified, usage } = params;
+    const ctx = { taskId };
+
+    // Metering an `exact` requirement throws in the SDK — the signature binds
+    // it to one value — so the option is only ever passed for `upto`.
+    let settleOpts: { amountAtomic?: string } | undefined;
+    if (this.pricing.scheme === 'upto') {
```
longfin: [P1] `classified.requirement`는 turn 1에 저장된 offering인데 여기의 `this.pricing`은 turn 2 시점의 live pricing입니다. Offering이 `upto(max=1000000)`인 동안 관리자가 `exact(10000)`로 repricing하면 이 branch가 실행되지 않아 SDK가 stored `upto` requirement의 ceiling 1000000을 그대로 settle합니다. 반대 방향은 작업 완료 뒤 metered-exact 오류로 실패합니다. 실제로 전액 settle되는 것을 재현했습니다. Offering 생성 당시 scheme/rates/minAmount를 함께 snapshot하고 settlement는 그 snapshot만 사용하거나, 가격 버전이 달라졌으면 새 offering을 발급해야 합니다.

packages/server/src/x402/store.ts
```diff
@@ -0,0 +1,144 @@
...
+    `;
+    const row = rows[0];
+    return row ? reviveEntry(row.entry) : undefined;
+  }
+
+  async update(taskId: string, patch: X402StoreEntryPatch): Promise<void> {
+    // Read-modify-write rather than a JSONB merge: the patch carries nested
+    // Date values that would have to be hand-serialized into a jsonb_set
+    // chain anyway, and the round-trip stays inside one statement's worth of
+    // latency. Concurrent patches for one task don't occur — a task's
```
longfin: [P1] 이 동시성 가정은 HTTP retry/resubmission에서는 성립하지 않습니다. 현재 SDK `classify()`는 stored entry가 이미 `verified`인지 확인하지 않으므로, 동일한 signed `payment-submitted`를 settlement 전에 두 번 보내면 두 요청 모두 `proceed`하고 verify도 두 번 호출됩니다. 이후 executor가 기존 binding을 밀어내기 전에 각각 backend 작업을 시작할 수 있어 한 번의 authorization으로 작업을 반복 실행시킬 수 있습니다. DB에서 `offered -> claimed/verified`를 조건부 UPDATE로 원자적으로 전이하고, 이미 claimed/verified인 submission은 forwarding 전에 거절해야 합니다.

packages/server/src/x402/store.ts
```diff
@@ -0,0 +1,144 @@
...
+  async delete(taskId: string): Promise<void> {
+    await this.sql`DELETE FROM x402_offerings WHERE task_id = ${taskId}`;
+  }
+
+  /**
+   * Drop rows whose expiry has passed. Lazy expiry already hides them from
+   * `get`, so this only reclaims space for offerings whose task was never
+   * completed or cancelled. Returns the number of rows removed.
+   */
+  async sweepExpired(): Promise<number> {
```
longfin: [P2] `sweepExpired()`의 호출부가 저장소 전체에 없습니다. TTL은 `get()`에서 expired row를 숨길 뿐 삭제하지 않으므로, caller가 turn 1의 `input-required`에서 떠날 때마다 row가 영구 누적됩니다. 공개 paid agent에는 인증 없이 반복 생성할 수 있어 DB growth 경로가 됩니다. 기존 hourly transient cleanup에 expired offering 삭제를 포함하거나 DB-native scheduled cleanup을 추가해야 합니다.

packages/server/src/admin-api.ts
```diff
@@ -219,6 +225,104 @@ export async function listCallers(
...
+      WHERE id = ${agentId}
+      RETURNING id AS agent_id
+    `;
+  });
+  // RLS hides agents the operator doesn't own, so zero rows is
+  // indistinguishable from "no such agent" — and deliberately so, since
+  // distinguishing them would confirm the existence of someone else's agent.
+  if (rows.length === 0) throw new AdminApiError('Agent not found or not authorized.', 404);
+
+  registry.updateX402Pricing(agentId, parsed);
```
longfin, line 305: [P2] 이 hot reload는 현재 프로세스의 in-memory registry만 갱신합니다. PR 설명처럼 여러 Fly instance가 있는 경우 admin PUT/DELETE를 처리한 instance와 agent WS를 보유한 instance가 다를 수 있어, 실제 agent는 재연결 전까지 이전 가격으로 계속 요구/정산합니다. 특히 clear 이후에도 다른 instance가 계속 과금할 수 있습니다. Postgres LISTEN/NOTIFY나 pub/sub로 invalidate하거나, turn 시작 시 DB의 pricing version을 확인해야 합니다.

packages/server/src/http.tsx
```diff
@@ -425,6 +429,60 @@ export function createHttpApp(opts: ServerHttpOptions): Hono {
...
+    const auth = await authOwnerSession(c);
+    if (!auth.ok) return adminApiUnauthorized(c, auth);
+    try {
+      return c.json(await getX402Pricing(opts.db, auth.principalId, c.req.param('id')));
+    } catch (err) {
+      return adminApiErrorResponse(c, err);
+    }
+  });
+
+  app.put('/admin-api/agents/:id/x402', async (c) => {
```
longfin, line 450: [P3] 새 endpoint는 PUT인데 위 전역 CORS 설정의 `allowMethods`는 `GET`, `POST`, `DELETE`, `OPTIONS`만 포함합니다. CLI의 Node fetch는 통과하지만 Vite dev UI 같은 cross-origin browser client는 preflight에서 차단됩니다. `PUT`을 허용 목록에 추가하고 OPTIONS 테스트를 넣어 주세요.

packages/server/src/x402/store.ts
longfin: Fixed in 11beaaa. The key is now `(agent_id, task_id)` and `get` / `update` / `delete` / the new `claim` and pricing accessors all filter on `this.agentId`; `ON CONFLICT` targets the composite. You were right that the column was already there — I had written it off in a comment as a label for operational queries when it was the scoping key.

Continuation tasks are the other half and I have filed them separately as #453: the same shared-store resolution also lets `tasks/get` and `tasks/cancel` reach another agent, which is an authorization gap independent of payments. Scoped-offering alone closes the money path, since B's gate now finds nothing for a task offered by A.

Covered by `src/x402/store.test.ts` against a real Postgres — cross-agent `get`/`getPricing`/`claim`/`delete` all refused, and the same taskId holding an independent offering per agent (which the old single-column PK silently collapsed).

packages/server/src/x402/gate.ts
longfin: Fixed in 11beaaa. `requestPayment` now writes a pricing snapshot alongside the offering, `open` hands it back, and `settle` takes it as a required parameter — the live `this.pricing` is no longer reachable from the settlement path at all.

I also took the belt-and-braces half of your suggestion: the metering branch requires `requirementScheme(classified.requirement) === 'upto'` as well as the snapshot agreeing, so no single stale value can send a metered amount at an exact signature or vice versa. A snapshot that is somehow missing refuses the call rather than falling back to live pricing, which is the bug itself.

Regression test in `gate.test.ts`: an `upto` offering, then a gate rebuilt with `exact` pricing for turn 2, asserting the facilitator sees the metered 450000 rather than the 1000000 ceiling.

packages/server/src/x402/store.ts
longfin: Fixed in 11beaaa, and you were right that the comment asserting single-in-flight was the bug. I confirmed `classify()` never reads `entry.status` — it checks existence, version, requirement match and payload shape, then returns valid.

Accepting a submission is now a conditional `UPDATE ... SET claimed_at = now() WHERE claimed_at IS NULL` on a new column, taken **before** verify so a replay costs no facilitator call and never reaches `sendToAgent`. I kept it on its own column rather than transitioning the entry's status, so the SDK's own lifecycle writes (`verified`, `completed`) cannot clobber it — `put` leaves both `claimed_at` and `pricing` alone on conflict.

The claim is deliberately never released. Releasing on a failed verify would restore the window, and nothing has been charged at that point, so the payer starting a new task is the cheaper failure.

Tested both ways in `store.test.ts`: sequential replay loses, and eight concurrent `claim()` calls produce exactly one winner.

packages/server/src/x402/store.ts
longfin: Fixed in 11beaaa. Promoted to a module-level `sweepExpiredX402Offerings(sql)` (it is a cross-agent job, not a per-store one) and wired into the existing hourly `cleanupExpiredTransients`, alongside the device-session and SIWE-nonce sweeps. Emits `x402_offerings_swept` when it removes anything. Test asserts it reclaims the lapsed row and leaves live ones alone.

packages/server/src/http.tsx
longfin: Fixed in 11beaaa — `PUT` added to `allowMethods`, with a comment noting why it is easy to miss (the CLI's Node fetch sends no preflight, so only browser clients hit it).

packages/server/src/admin-api.ts
longfin: Fixed in 11beaaa, via Postgres `LISTEN`/`NOTIFY` rather than a per-turn DB read — the write is rare and the read is on every paid call.

`setX402Pricing` / `clearX402Pricing` patch the local registry as before and then `NOTIFY x402_pricing_changed` with the agent id. Each instance subscribes at startup (`x402/pricing-watch.ts`); on receipt it drops the notification immediately unless it actually holds that agent's socket, then re-reads the row and calls `registry.updateX402Pricing`. It re-reads rather than trusting the payload so two racing writes cannot leave an instance holding whichever message arrived last.

Failure posture matches the rest of the path: a notify that fails is logged and does not fail the admin write (which has already committed), a subscription that cannot be established is logged and the deployment degrades to the previous reconnect-scoped behavior, and a row that fails to parse makes the agent free rather than taking it offline. The clear direction you flagged is the one this most protects.

Docs updated — the previous "hot-reloaded, no reconnect" claim was only true single-instance.

packages/server/src/executor.ts
```diff
@@ -325,18 +496,27 @@ export class WSForwardingExecutor extends AgentExecutor {
         }
         // status event
         if (event.final) sawFinalEvent = true;
-        history = appendHistoryMessage(history, event.status.message);
-        task.history = history;
         if (TERMINAL_STATES.has(event.status.state)) {
+          // Settle before recording anything: a settlement failure replaces
+          // the agent's terminal event outright, and appending the original
+          // first would leave both in history.
+          const terminal =
```
longfin, line 531: [P1] Settlement 실패 전에 결과물이 이미 caller에게 전달됩니다. artifact는 위에서 즉시 `yield`되고, 여기서 terminal event를 받은 뒤에야 settle합니다. verify 후 payer가 잔액을 이동하거나 Permit을 무효화해 settle을 실패시키면 caller는 전체 스트리밍 출력물을 받은 뒤 `failed` 상태만 받습니다. 비스트리밍도 아래에서 `accumulatedArtifacts`를 task에 붙이므로 결과가 반환됩니다. `upto` 결과를 settlement 성공까지 버퍼링한 뒤 방출하거나, 이 미결제 작업 유출을 명시적으로 수용하는 설계가 필요합니다.

packages/server/src/x402/gate.ts
```diff
@@ -0,0 +1,514 @@
...
+        if (metadata === undefined) {
+          // `requestPayment` always yields a `request-input`; reaching here
+          // means the SDK contract changed under us. Refuse rather than
+          // silently serve the call for free.
+          throw new Error('x402 requestPayment yielded no request-input event');
+        }
+        // Freeze the terms alongside the offering. Turn 2 settles from this,
+        // so a reprice between the turns cannot change what the payer is
+        // charged for a requirement they already signed.
+        await this.sidecar.putPricing(taskId, this.pricing);
```
longfin: [P1] Offering 본문과 pricing snapshot이 서로 다른 SQL로 저장되어 원자적이지 않습니다. 같은 taskId의 중복 turn-1이 repricing과 겹치면 `old requestPayment.put -> new requestPayment.put -> new putPricing -> old putPricing` 순서로 DB에 `upto requirement + exact snapshot`이 남을 수 있습니다. 그 `upto` 서명은 classify/verify를 통과하지만 settle 쪽은 scheme 불일치를 거절하지 않고 metering만 생략하여 SDK가 authorized ceiling을 그대로 settle합니다. Offering과 snapshot을 한 원자적 write로 설치하고, snapshot scheme과 signed requirement scheme이 다르면 반드시 fail-closed 해야 합니다.

packages/server/src/agent-card.ts
```diff
@@ -119,6 +130,48 @@ export function buildAgentA2XServer(
...
+  // side that knows whether a payment gate is actually installed, and pricing
+  // is DB-owned rather than declared by the connecting client.
+  //
+  // `required: false` — a caller that cannot pay still gets a well-formed
+  // `input-required` response naming the price, which is more useful than
+  // being refused at extension activation. `params` mirrors the offering so a
+  // client can decide whether it is willing to pay before spending a turn.
+  if (conn.x402Pricing) {
+    const pricing = conn.x402Pricing;
+    a2xServer.addExtension({
```
longfin, line 174: [P2] 주석과 달리 bridge가 x402 advertisement를 독점하지 않습니다. 위 loop가 wire-declared x402 URI를 그대로 `addExtension`한 뒤 여기서 DB 기반 항목을 append합니다(`addExtension`은 append-only). 그래서 free agent도 임의의 x402를 광고할 수 있고, paid agent는 같은 URI가 서로 다른 가격으로 중복되며, legacy V1 URI와 V2 gate도 함께 노출될 수 있습니다. Wire card에서 모든 x402 URI를 먼저 제거한 뒤 DB 기반 항목만 추가해야 합니다.

packages/server/src/agent-card.ts
```diff
@@ -36,7 +41,13 @@ export function buildAgentA2XServer(
     ? `${opts.publicUrl}/agents/${conn.agentId}`
     : `/agents/${conn.agentId}`;
 
-  const executor = new WSForwardingExecutor(conn.agentId, registry, taskStore);
+  const executor = new WSForwardingExecutor(
+    conn.agentId,
+    registry,
+    taskStore,
+    undefined,
+    opts.db ? { sql: opts.db, resource: url } : undefined,
```
longfin: [P2] `PUBLIC_URL`이 없으면 위 `url`은 `/agents/:id` 상대경로인데 이것이 그대로 x402 `resource`가 됩니다. `pricingToAccepts`의 계약대로 strict facilitator는 non-URL resource를 거절하므로, PUBLIC_URL 없이 기동 가능한 배포에서는 유료 agent가 offering은 만들지만 verify/settle이 실패합니다. 유료 pricing 사용 시 absolute PUBLIC_URL을 필수로 검증하거나 안전하게 절대 URL을 구성해야 합니다.

longfin (commented):
최신 HEAD를 추가 검토했습니다. 결제 완료 보장과 pricing snapshot 원자성에 merge 전 확인이 필요한 P1 2건, AgentCard 신뢰 경계와 resource URL 관련 P2 2건을 남깁니다. 작성자 계정이라 REQUEST_CHANGES 대신 COMMENT review로 제출합니다.

packages/server/src/x402/gate.ts
longfin: Fixed in 19b921d, and you were right that my "belt and braces" was the bug rather than the guard. Requiring both the snapshot and the signed requirement to say `upto` meant a disagreement fell through to the **un-metered** branch — which hands the SDK no amount, and for an `upto` requirement that settles the full authorized ceiling. Skipping the meter is not a safe default.

A mismatch now fails closed with `x402_scheme_mismatch`, before verify, so it costs no facilitator call either.

On atomicity: the SDK owns the entry write, so the two genuinely cannot share one statement. I inverted the order instead — the snapshot is written first (as an upsert), so an entry never exists without terms. A crash between them leaves terms with no offering, which reads as "no offering" and is refused; the reverse leaves an offering nothing can price. The placeholder row that creates is hidden from `get` and `claim` by a `jsonb_exists(entry, 'accepts')` filter rather than being allowed to look like an empty offering.

That still does not make the pair atomic under the interleaving you described, so the fail-closed check is the real defense and the ordering is depth. Regression test in `gate.test.ts` publishes an `upto` offering, corrupts the snapshot to `exact`, and asserts the call is refused with no facilitator contact.

packages/server/src/agent-card.ts
longfin: Fixed in 19b921d. All x402 URIs are now stripped from the wire card before the DB-owned entry is appended — using the SDK's `isX402ExtensionUri`, so the legacy V1 URI is caught too, not just the foundation one. Dropped for free agents as well as paid, and logged as `x402_wire_advertisement_dropped`.

The comment claiming the bridge owned this advertisement is now true rather than aspirational. Three tests: a free agent advertises nothing while unrelated wire extensions still pass through, a paid agent has exactly one entry carrying the DB `payTo` rather than the wire-declared one, and the no-PUBLIC_URL case below.

packages/server/src/agent-card.ts
longfin: Fixed in 19b921d. Payments are now wired only when `publicUrl` is set, so a paid agent without it serves for free and logs `x402_config_invalid` naming the reason. The card quotes no price either — advertising one the gate is not installed to collect is the same failure in a different place.

Chose withholding over hard-failing startup to match the posture already used for an unparseable pricing row: a working deployment with a visible misconfiguration beats an outage, and the direction of the error is "not charging" rather than "charging into a loop that cannot settle". Documented under Deployment settings.

packages/server/src/executor.ts
longfin: Confirmed and **accepted rather than fixed** — with the reasoning now in the code and the docs instead of being implicit, which is the part that was genuinely missing.

You are right about the mechanics: artifacts `yield` as they arrive, settlement cannot begin until the terminal event, and the non-streaming path attaches `accumulatedArtifacts` to the task regardless. A payer who invalidates between verify and settle gets the output and a `failed` status.

Both ways of closing it are worse for this product. Buffering until settlement lands removes streaming from exactly the agents people pay for. Settling before the work charges for tasks that fail — and on a bridge fronting coding agents, timeouts and backend errors are ordinary, while an adversarial payer is not. Undercharging in a rare attack beats overcharging on routine failure, so the current order is the one I want.

Under `upto` there is additionally nothing to meter before the work finishes, so settle-first is not even available there.

What I changed: the trade-off is stated in `settleTerminal`'s docstring and in a dedicated docs section, and settlement failure after delivery now emits `x402_unpaid_delivery` so the exposure is countable rather than invisible. Flagged in the docs that a rising rate is worth investigating and a flat zero is the expected state.

Happy to revisit if you would rather buffer for `exact` specifically — that one *could* settle before forwarding, since the amount is fixed and known up front. It would mean charging for failed tasks, which is why I did not.

longfin: Done in c297075 — `exact` now settles inside the gate, after verify and before `task.assign`.

A payment that cannot be collected halts the turn: no charge, no work, backend never contacted. That is strictly better than the old order for the case you raised, where the caller kept the output.

The accepted cost is the mirror: a task that fails after payment is a real loss to the payer, since we cannot refund. `x402_paid_task_failed` counts it, and the receipt is attached to the failed task so they hold proof of what they paid.

`upto` is unchanged and cannot change — the charge is metered from consumption, which does not exist until the work is done. Its window stays, now documented as inherent to metered billing rather than as a choice, and still counted by `x402_unpaid_delivery`.

Two structural notes from doing it:

- The gate outcome now names which happened (`settled` / `deferred`) instead of the executor re-deriving it from pricing, so the settle-first and settle-last paths cannot be confused.
- `settle` no longer clears the offering. Under `exact` it runs *before* the work, and dropping the row there would release the claim mid-task — a resubmission would then be treated as a fresh payment and charged again. The executor clears at the terminal on every path instead. This is the one place where the new order needed the claim from the earlier round to already be in place.

Tests updated to the new invariants: a flat fee settles during `open` and hands back the receipt, an uncollectable one halts before the backend, and a metered offering asserts `settleCalls === 0` at proceed time.

packages/server/src/x402/gate.ts
```diff
@@ -0,0 +1,589 @@
...
+
+      // The snapshot and the requirement the payer actually signed must name
+      // the same scheme. Concurrent turn-1s racing a reprice can leave the two
+      // disagreeing, and the disagreement is expensive in one specific
+      // direction: an `upto` requirement paired with an `exact` snapshot skips
+      // the metering branch, which hands the SDK no amount and settles the
+      // full authorized ceiling. Refuse rather than let a mismatch pick a
+      // price — merely skipping the metering is not a safe default here.
+      const signedScheme = requirementScheme(classified.requirement);
+      if (signedScheme !== settlementPricing.scheme) {
```
longfin, line 330: [P1] 이 검사는 scheme만 비교하므로 같은 scheme 안에서 offering과 pricing snapshot이 엇갈리는 경합을 막지 못합니다. 예를 들어 두 `upto` turn-1이 repricing과 겹쳐 A:`putPricing(old)`, B:`putPricing(new)`, B:`requestPayment.put(new)`, A:`requestPayment.put(old)` 순서가 되면 DB에는 A의 offering과 B의 rates/minAmount snapshot이 남습니다. 둘 다 `upto`라 이 guard를 통과하고, A가 서명한 cap 안에서라도 B의 단가로 과금될 수 있습니다. Offering 본문과 pricing을 원자적으로 저장하거나, offering에 terms fingerprint/version을 결합해 전체 조건이 정확히 일치하지 않으면 fail-closed 해야 합니다.

packages/server/src/x402/gate.ts
```diff
@@ -0,0 +1,589 @@
...
+      // Safe to do here only because `claim` above is a one-shot: without it a
+      // replayed submission would settle a second time.
+      const settled = await this.settle({
+        taskId,
+        contextId,
+        classified,
+        pricing: settlementPricing,
+      });
+      if (settled.kind === 'failed') return { kind: 'halt', event: settled.event };
+      return { kind: 'proceed', settlement: { mode: 'settled', metadata: settled.metadata } };
```
longfin, line 393: [P1] 여기서 `exact`가 이미 settle된 뒤에도 executor의 `sendToAgent()`가 false가 될 수 있습니다(verify/settle 중 agent disconnect). 현재 `!sent` branch는 실패 event를 그대로 yield해 `settleTerminal()`을 거치지 않으므로, 결제된 실패 task에 receipt가 붙지 않고 `x402_paid_task_failed`도 기록되지 않으며 offering도 terminal 시점에 clear되지 않습니다. `!sent` 실패도 task/history를 갱신하기 전에 `settleTerminal(gate, settlement, ...)`로 정규화해야 합니다(`deferred`의 미청구 offering 정리도 함께 해결됩니다).

packages/server/src/x402/store.ts
```diff
@@ -0,0 +1,240 @@
...
+    // `entry` gets a placeholder on insert and is immediately overwritten by
+    // the SDK's `put`. It is never read in that state: `get` goes through the
+    // SDK's shape and the gate only reaches settlement via `classify`, which
+    // requires a real entry.
+    await this.sql`
+      INSERT INTO x402_offerings (task_id, agent_id, entry, pricing, updated_at)
+      VALUES (
+        ${taskId},
+        ${this.agentId},
+        '{}'::jsonb,
```
longfin: [P2] 이 placeholder는 `expires_at` 없이 삽입됩니다. 여기와 SDK의 후속 `put()` 사이에서 프로세스가 종료되거나 `requestPayment()`가 예외를 던지면 `entry='{}'`, `expires_at=NULL`인 row가 남고, 현재 sweeper는 `expires_at IS NOT NULL`인 row만 지우므로 영구 누적됩니다. Placeholder에도 임시 TTL을 넣거나 `jsonb_exists(entry, 'accepts') = false`인 오래된 row를 `updated_at` 기준으로 sweep하고, 후속 write 실패 경로를 테스트해 주세요.

longfin (commented):
최신 HEAD `c297075`를 다시 검토했습니다. 앞선 AgentCard/PUBLIC_URL 수정과 `exact` 선결제 방향은 확인했지만, merge 전에 추가로 확인해야 할 P1 2건과 P2 1건이 남아 있습니다. 작성자 계정이라 REQUEST_CHANGES 대신 COMMENT review로 제출합니다.

packages/server/src/x402/gate.ts
longfin: Fixed in 2f25252, at the root rather than by widening the comparison. You are right that the scheme check was insufficient — and worse than I realized, because `rates` and `minAmount` are ours and never appear in the x402 offering at all, so no comparison against `accepts` or the signed requirement could have caught your example either.

The terms now ride in the **same INSERT** as the offering. The SDK owns that write, so the pricing travels through `AsyncLocalStorage` — per-call context, which is the point: a module-level variable would have exactly the race we are trying to remove. `put` checks the context's taskId against the entry's before applying, so a context can never leak onto a different offering, and the SDK's later lifecycle patches run outside the context and leave the frozen terms alone via `COALESCE`.

The scheme guard stays as depth, but it is no longer load-bearing.

Test in `store.test.ts` runs two `publishing()` calls for one task concurrently with different amounts and asserts the stored terms are the ones that produced the stored offering, whichever won.

packages/server/src/x402/gate.ts
longfin: Fixed in 2f25252. The `!sent` branch now routes through `settleTerminal(gate, settlement, failEvent, ...)` before task/history are updated, so it is normalized like every other terminal: the receipt is attached, `x402_paid_task_failed` fires, and the offering is released. As you noted, that also cleans up the uncharged `deferred` offerings that were leaking on this path.

Good catch that this only became a real exposure with the previous commit — before `exact` settled up front, the worst case here was a stale offering row rather than a paid task with no receipt.

packages/server/src/x402/store.ts
longfin: Resolved by the atomicity fix in 2f25252 rather than by sweeping. The placeholder only existed to let the terms be written before the offering; now that both go in one INSERT there is no in-between state, so no `expires_at IS NULL` row can be left behind. `getPricing` reads the column off the real row, and the `jsonb_exists(entry, 'accepts')` filters that guarded against the placeholder are gone with it.

You were right that the sweeper could never have reclaimed those rows — worth noting it still only handles `expires_at IS NOT NULL`, which is now correct because every row the store writes carries the offering TTL.

PR #472, opened 2026-08-20
fix(claude): preserve completed result after narrated retry

### LINKED ISSUE #471: narrated-tool-call 재시도가 정상 완료된 태스크를 하드 실패로 뒤집음 — 발동 8건 중 8건 실패 (성공 0)
## 요약

`--claude-retry-narrated-tool-call` (#447, #441의 완화책)이 **정상적으로 완료된 태스크를 하드 실패로 뒤집고 있습니다.**

`vicoop-provider` 운영 호스트의 저널 전체 보존 구간(2026-06-10 ~ 2026-08-19)에서 이 재시도는 **8회 발동해 8회 모두 `task.fail`로 끝났습니다. 성공 사례가 0건입니다.**

그중 마지막 1건(`2026-08-19T17:09:34Z`)이 라우터의 24시간 쿨다운을 유발해 Claude 계열 모델 전체를 내렸습니다 → planetarium/vicoop-router#152.

### 1. 첫 시도는 정상 완료된다

모델이 완결된 최종 보고서를 산문으로 내놓고, tool call은 0건입니다. 이건 **정상**입니다 — 더 호출할 도구가 없어서 보고로 끝낸 케이스입니다. 이 시점의 상태는 `sawCompletedResult = true`, `sawErrorResult = false`.

### 2. `shouldRetryNarratedToolCall`이 오탐한다

[`claude.ts:505`](packages/client/src/backends/claude.ts#L505)의 판정 기준은 **"본문이 등록된 caller tool 이름을 언급하는가"** 하나입니다.

문제는 실패한 태스크들의 본문이 하나같이 **"그 도구가 동작하지 않았다"고 설명하는 보고서**였다는 점입니다. 실제 `9f7cdc53`의 본문 일부:

> I've exhausted the available lookup paths for this follow-up — both `msu-resource_get_sprite_data` and `maple-lookup_get_sprite_data` reject `category: "map"` ... and the `bash` tool's external-directory permission is hard-denied in this session

`msu-resource_get_sprite_data`, `maple-lookup_get_sprite_data`, `bash` — 전부 등록된 도구 이름입니다. 휴리스틱은 **"X를 호출하겠다"와 "X가 동작하지 않는다"를 구분할 수 없습니다.** 후자는 도구가 막힌 상황에서 나오는 가장 자연스러운 최종 보고 형태이므로, 오탐이 아니라 구조적으로 걸릴 수밖에 없는 패턴입니다.

코드 주석은 이 위험을 이미 정확히 예견하고 있습니다:

> Tuned for PRECISION, not recall. A false positive costs a wasted turn AND asks a model that legitimately finished to invent a tool call it does not need, which is worse than missing a stall.

실측 결과 정밀도는 **0/8**이었습니다.

### 3. 재시도는 반드시 실패한다

재시도는 [`claude.ts:3243`](packages/client/src/backends/claude.ts#L3243)에서 `--session-id`만 `--resume`로 바꾸고 **나머지 인자를 그대로 유지합니다 — `--max-turns 1` 포함.**

호출할 도구가 애초에 없는 상황이므로 모델은 nudge를 받고도 tool call을 만들 수 없고, 턴은 tool call 없이 끝나 claude가 exit 1 + `terminal_reason: "max_turns"` / `is_error: true`로 종료합니다. 실제 로그:

```
claude exited with code 1 [stdout: ...,"terminal_reason":"max_turns",...,"errors":["Reached maximum number of turns (1)"]}]
```

즉 오탐으로 발동한 재시도는 **구조적으로 실패가 예정돼 있습니다.**

### 4. 재시도의 실패가 첫 시도의 성공을 덮어쓴다

핵심입니다.

- `sawCompletedResult` / `sawErrorResult`는 [`claude.ts:2657`](packages/client/src/backends/claude.ts#L2657)에서 선언되어 **재시도 블록 바깥에 있습니다.**
- 재시도는 `attachStreams(child)`로 핸들러만 새 자식에 다시 붙일 뿐 **이 플래그들을 초기화하지 않습니다.**
- 따라서 재시도의 `is_error: true` result가 [`claude.ts:3066`](packages/client/src/backends/claude.ts#L3066)에서 `sawErrorResult = true`를 래치합니다.
- [`claude.ts:3330`](packages/client/src/backends/claude.ts#L3330)의 `treatExitAsSuccess`는 `!sawErrorResult`를 요구하므로 false가 되고,
- `exit`도 이미 재시도 자식의 것으로 교체된 상태라 → **하드 `task.fail code=claude_exit_nonzero`.**

실패 메시지에는 `reasonText = finalText`가 그대로 쓰여서, **첫 시도가 만든 멀쩡한 보고서 본문이 "에러 메시지"로 라우터에 전달됩니다.** vicoop-router의 `lastErrorMessage`에 보고서 전문이 박혀 있는 이유입니다.

### 5. 의도한 불변식과의 모순

[`claude.ts:3266`](packages/client/src/backends/claude.ts#L3266) 주석은 이 불변식을 명시하고 있습니다:

> A failed retry must not fail the task: the first attempt already produced a legitimate (if useless) completed result, and surfacing a spawn error here would turn a degraded turn into a hard failure.

그러나 이 보호는 `try`/`catch`가 잡는 **spawn 에러에만** 적용됩니다. 실제로 발생하는 경로인 **재시도 자식의 비정상 종료 코드**는 catch를 타지 않고 그대로 아래 exit 처리로 흘러갑니다. 의도한 불변식이 실제 발동 경로에서 깨져 있습니다.

## 실측

`vicoop-provider` / `vicoop-bridge-claude.service`, 저널 전 구간:

| 지표 | 값 |
|---|---|
| `caller-tool turn produced no tool calls` 관측 | 76 |
| 재시도 발동 | **8** |
| 재시도 → `task.fail` | **8 (100%)** |
| 재시도 → 성공 | **0** |
| `narrated-tool-call retry failed` (spawn 실패, 무해) | 0 |

발동 8건 전부:

```
2026-08-10T05:18:37Z  d708e06b-2f94-4747-aad2-f264c88fac9c  -> task.fail
2026-08-10T12:19:44Z  5fb8a6b8-91d3-4671-ac24-22beb7b35065  -> task.fail
2026-08-10T12:30:24Z  77132f9e-1170-4f7c-8c15-f77170da0929  -> task.fail
2026-08-18T07:39:24Z  185e6d59-b208-485a-b5f2-655acffaad33  -> task.fail
2026-08-18T11:27:31Z  aee4ff8f-3d49-4a68-96e0-d97fd35d943f  -> task.fail
2026-08-19T16:56:19Z  545e835e-ba5d-4799-91d5-b9cc1ac89d97  -> task.fail
2026-08-19T16:59:31Z  8d97e447-8d91-4a1b-9e32-663ed35df6e0  -> task.fail
2026-08-19T17:09:34Z  9f7cdc53-5bb4-4dde-a7b2-4a12c76f1541  -> task.fail   ← 라우터 쿨다운 유발
```

재시도가 없었다면 이 8건은 전부 `treatExitAsSuccess` 두 번째 절(`sawCompletedResult && !sawErrorResult && stderr 비어있음`)에 걸려 **정상 완료로 처리됐을 것**입니다. 즉 이 기능은 순수하게 손실만 만들고 있습니다.

## 영향

1. **호출자가 결과물을 잃습니다.** 완결된 보고서가 담긴 태스크가 실패로 반환됩니다.
2. **라우터 쿨다운 연쇄.** `agent_error`가 vicoop-router에서 `misconfigured`로 분류되어 24시간 쿨다운 → 해당 에이전트가 유일 멤버인 풀이 통째로 다운됐습니다 (planetarium/vicoop-router#152). 2026-08-20 02:09 KST 실제 장애로 이어졌습니다.
3. 운영 호스트 `vicoop-provider`의 `ExecStart`에 이 플래그가 켜져 있습니다:
   ```
   vicoop-client start --backend claude ... --claude-retry-narrated-tool-call
   ```

## 제안하는 수정

우선순위 순:

1. **재시도 실패가 태스크를 죽이지 않게 한다 (가장 시급).** 첫 시도의 `sawCompletedResult` / `sawErrorResult` / `exit`을 재시도 전에 스냅샷해 두고, 재시도가 tool call을 만들어내지 못하면 **첫 시도 결과로 되돌린다.** [`claude.ts:3266`](packages/client/src/backends/claude.ts#L3266) 주석이 선언한 불변식을 spawn 에러뿐 아니라 비정상 종료까지 확장하는 것입니다. 이 하나만으로 8/8 실패가 전부 정상 완료로 바뀝니다.
2. **재시도에 `--max-turns 1`을 그대로 쓰지 않는다.** nudge 턴은 tool call을 만들라는 요구이므로 최소 2턴이 필요합니다. 현재 설정에서는 성공 자체가 불가능한 구조입니다.
3. **휴리스틱 정밀도 개선.** "도구 이름 언급"만으로는 부족합니다. 최소한 부정 문맥(`reject`, `denied`, `not available`, `NOT FOUND`, `UNABLE TO`, `failed`)이나 최종 보고 형태(`## RESULT`, 표, 대용량 textLen)를 배제 신호로 넣어야 합니다. 다만 1번이 적용되면 오탐 비용이 "낭비된 턴 1회"로 떨어지므로 긴급도는 내려갑니다.
4. **플래그를 기본 비활성 유지 + 운영 호스트에서 잠정 해제 검토.** 현재 실측 기대값이 순손실이므로, 1·2번이 반영되기 전까지 `vicoop-provider`에서 내리는 것이 안전합니다.

## 재현

1. caller-tool dispatch(`--max-turns 1`)로 태스크를 보내고, `--claude-retry-narrated-tool-call`을 켠다.
2. 모델이 **등록된 도구 이름을 언급하면서 그 도구가 실패했다고 설명하는** 최종 보고서를 tool call 없이 내도록 유도한다 (예: 존재하지 않는 category를 요구해 도구가 거부하게 만든 뒤 결과 보고를 시킨다).
3. `retrying narrated tool call` 로그가 뜨고, 이어서 `task.fail code=claude_exit_nonzero`로 끝나는지 확인한다.
4. 실패 메시지가 1번에서 만들어진 보고서 본문 그대로인지 확인한다.

## 관련

- #441 claude/fable: turn ends after announcing work, emitting no tool calls (openai-compat) — 이 재시도가 완화하려던 원 현상
- #447 feat(claude): optionally retry a turn that narrated its tool call — 도입 PR (`9543554`)
- planetarium/vicoop-router#152 — 이 실패가 라우터에서 24h 쿨다운으로 증폭된 건

longfin:
## Summary

- treat the narrated-tool retry as speculative and keep the first completed result unless the retry captures a new caller tool call
- restore terminal flags, diagnostics, response text, and model identity when the retry produces no caller tool call
- strengthen the retry-success test with a real caller-tools MCP invocation and production-like exit code 1
- add an exact regression for completed → max_turns/is_error → exit 1

Fixes #471.

## Verification

- `pnpm --filter @vicoop-bridge/client typecheck`
- `pnpm --filter @vicoop-bridge/client build`
- `pnpm --filter @vicoop-bridge/client test` — 895 passed
- `pnpm changeset status`

### DIFF
--- a/packages/client/src/backends/claude.test.ts
+++ b/packages/client/src/backends/claude.test.ts
@@ -6127,18 +6127,27 @@ function twoAttemptSpawn(
 ): FakeSpawn & {
   argvs: string[][];
   stdins: string[][];
+  assertAttemptHooksSucceeded: () => void;
 } {
   const argvs: string[][] = [];
   const stdins: string[][] = [];
+  let attemptHookError: Error | null = null;
   let n = 0;
   const base = makeFakeSpawn((child) => {
     const attempt = n++;
     const lines = scripts[Math.min(attempt, scripts.length - 1)] ?? [];
     const exitCode = exitCodes[Math.min(attempt, exitCodes.length - 1)] ?? 0;
-    setImmediate(async () => {
-      await beforeAttempt?.(attempt);
-      for (const l of lines) child.emitStdout(l.endsWith('\n') ? l : `${l}\n`);
-      setImmediate(() => child.finish(exitCode, null));
+    setImmediate(() => {
+      void Promise.resolve()
+        .then(() => beforeAttempt?.(attempt))
+        .then(() => {
+          for (const l of lines) child.emitStdout(l.endsWith('\n') ? l : `${l}\n`);
+          setImmediate(() => child.finish(exitCode, null));
+        })
+        .catch((err: unknown) => {
+          attemptHookError = err instanceof Error ? err : new Error(String(err));
+          child.finish(1, null);
+        });
     });
   });
   const wrapped = {
@@ -6151,8 +6160,15 @@ function twoAttemptSpawn(
     },
     argvs,
     stdins,
+    assertAttemptHooksSucceeded() {
+      if (attemptHookError) throw attemptHookError;
+    },
   };
-  return wrapped as FakeSpawn & { argvs: string[][]; stdins: string[][] };
+  return wrapped as FakeSpawn & {
+    argvs: string[][];
+    stdins: string[][];
+    assertAttemptHooksSucceeded: () => void;
+  };
 }
 
 const NARRATED = JSON.stringify({
@@ -6209,6 +6225,7 @@ test('narrated tool call: opt-in resumes once with the staged prompt still reada
   });
   const { emit, frames } = collect();
   await backend.handle(narratedToolCallTask(), emit, NEVER);
+  fake.assertAttemptHooksSucceeded();
 
   assert.equal(fake.argvs.length, 2, 'expected exactly one corrective retry');
   // Attempt 1 mints the session, attempt 2 continues it — same id either way.
--- a/packages/client/src/backends/claude.test.ts
+++ b/packages/client/src/backends/claude.test.ts
@@ -6120,7 +6120,11 @@ function narratedToolCallTask(): TaskAssignFrame {
 
 // Scripts each spawn separately so attempt 1 can narrate and attempt 2 emit a
 // real tool call, and records the argv of every spawn.
-function twoAttemptSpawn(scripts: readonly (readonly string[])[]): FakeSpawn & {
+function twoAttemptSpawn(
+  scripts: readonly (readonly string[])[],
+  exitCodes: readonly (number | null)[] = [],
+  beforeAttempt?: (attempt: number) => void | Promise<void>,
+): FakeSpawn & {
   argvs: string[][];
   stdins: string[][];
 } {
@@ -6128,11 +6132,13 @@ function twoAttemptSpawn(scripts: readonly (readonly string[])[]): FakeSpawn & {
   const stdins: string[][] = [];
   let n = 0;
   const base = makeFakeSpawn((child) => {
-    const lines = scripts[Math.min(n, scripts.length - 1)] ?? [];
-    n++;
-    setImmediate(() => {
+    const attempt = n++;
+    const lines = scripts[Math.min(attempt, scripts.length - 1)] ?? [];
+    const exitCode = exitCodes[Math.min(attempt, exitCodes.length - 1)] ?? 0;
+    setImmediate(async () => {
+      await beforeAttempt?.(attempt);
       for (const l of lines) child.emitStdout(l.endsWith('\n') ? l : `${l}\n`);
-      setImmediate(() => child.finish(0, null));
+      setImmediate(() => child.finish(exitCode, null));
     });
   });
   const wrapped = {
@@ -6175,11 +6181,32 @@ const REAL_CALL = JSON.stringify({
 const OK_RESULT = JSON.stringify({ type: 'result', subtype: 'success', terminal_reason: 'completed', result: '' });
 
 test('narrated tool call: opt-in resumes once with the staged prompt still readable', async () => {
-  const fake = twoAttemptSpawn([
-    [NARRATED, OK_RESULT],
-    [REAL_CALL, OK_RESULT],
-  ]);
-  const backend = createClaudeBackend({ spawn: fake.spawn, claudeRetryNarratedToolCall: true });
+  let invokeRetryTool: (() => Promise<void>) | null = null;
+  const fake = twoAttemptSpawn(
+    [
+      [NARRATED, OK_RESULT],
+      [REAL_CALL, OK_RESULT],
+    ],
+    [1, 1],
+    async (attempt) => {
+      if (attempt !== 1) return;
+      assert.ok(invokeRetryTool, 'caller-tools MCP must be ready before the retry');
+      await invokeRetryTool();
+    },
+  );
+  const backend = createClaudeBackend({
+    spawn: fake.spawn,
+    claudeRetryNarratedToolCall: true,
+    onCallerToolsMcpReady: (server) => {
+      invokeRetryTool = async () => {
+        await server.invokeForTest({
+          callId: 'toolu_1',
+          toolName: 'glob',
+          arguments: { pattern: '**/*' },
+        });
+      };
+    },
+  });
   const { emit, frames } = collect();
   await backend.handle(narratedToolCallTask(), emit, NEVER);
 
@@ -6203,6 +6230,43 @@ test('narrated tool call: opt-in resumes once with the staged prompt still reada
   assert.equal(frames.at(-1)?.type, 'task.complete');
   assert.equal(frames.some((frame) => frame.type === 'task.fail'), false);
   await assert.rejects(fs.stat(retryPromptPath), 'prompt file must be deleted after retry');
+});
+
+test('narrated tool call: a failed retry preserves the first completed result', async () => {
+  const finalReport = 'Final report: the registered `glob` tool rejected this request.';
+  const report = JSON.stringify({
+    type: 'assistant',
+    message: {
+      role: 'assistant',
+      model: 'claude-fable-5',
+      content: [{ type: 'text', text: finalReport }],
+    },
+  });
+  const completed = JSON.stringify({
+    type: 'result',
+    subtype: 'success',
+    terminal_reason: 'completed',
+    is_error: false,
+    result: finalReport,
+  });
+  const maxTurns = JSON.stringify({
+    type: 'result',
+    subtype: 'error_during_execution',
+    terminal_reason: 'max_turns',
+    is_error: true,
+    result: finalReport,
+    errors: ['Reached maximum number of turns (1)'],
+  });
+  const fake = twoAttemptSpawn([[report, completed], [maxTurns]], [1, 1]);
+  const backend = createClaudeBackend({ spawn: fake.spawn, claudeRetryNarratedToolCall: true });
+  const { emit, frames } = collect();
+
+  await backend.handle(narratedToolCallTask(), emit, NEVER);
+
+  assert.equal(fake.argvs.length, 2, 'expected exactly one corrective retry');
+  assert.equal(frames.some((frame) => frame.type === 'task.fail'), false);
+  assert.equal(frames.at(-1)?.type, 'task.complete');
+  assert.equal(textOf(frames.at(-1)!), finalReport);
 });
 
 test('narrated tool call: off by default — no retry, no extra spawn', async () => {
--- a/packages/client/src/backends/claude.test.ts
+++ b/packages/client/src/backends/claude.test.ts
@@ -6122,34 +6122,45 @@ function narratedToolCallTask(): TaskAssignFrame {
 // real tool call, and records the argv of every spawn.
 function twoAttemptSpawn(
   scripts: readonly (readonly string[])[],
-  exitCodes: readonly (number | null)[] = [],
-  beforeAttempt?: (attempt: number) => void | Promise<void>,
+  exitCodes: readonly number[] = [],
+  beforeAttempt?: (attempt: number, child: FakeChild) => void | Promise<void>,
 ): FakeSpawn & {
-  argvs: string[][];
-  stdins: string[][];
-  assertAttemptHooksSucceeded: () => void;
+  readonly argvs: string[][];
+  readonly stdins: string[][];
 } {
   const argvs: string[][] = [];
   const stdins: string[][] = [];
-  let attemptHookError: Error | null = null;
+  let attemptError: Error | null = null;
   let n = 0;
   const base = makeFakeSpawn((child) => {
     const attempt = n++;
     const lines = scripts[Math.min(attempt, scripts.length - 1)] ?? [];
-    const exitCode = exitCodes[Math.min(attempt, exitCodes.length - 1)] ?? 0;
+    const exitCode = exitCodes.length > 0
+      ? exitCodes[Math.min(attempt, exitCodes.length - 1)]!
+      : 0;
+    const failAttempt = (err: unknown): void => {
+      attemptError ??= err instanceof Error ? err : new Error(String(err));
+      child.finish(1, null);
+    };
     setImmediate(() => {
       void Promise.resolve()
-        .then(() => beforeAttempt?.(attempt))
-        .then(() => {
-          for (const l of lines) child.emitStdout(l.endsWith('\n') ? l : `${l}\n`);
-          setImmediate(() => child.finish(exitCode, null));
-        })
-        .catch((err: unknown) => {
-          attemptHookError = err instanceof Error ? err : new Error(String(err));
-          child.finish(1, null);
-        });
+        .then(() => beforeAttempt?.(attempt, child))
+        .then(
+          () => {
+            try {
+              for (const l of lines) child.emitStdout(l.endsWith('\n') ? l : `${l}\n`);
+              setImmediate(() => child.finish(exitCode, null));
+            } catch (err) {
+              failAttempt(err);
+            }
+          },
+          failAttempt,
+        );
     });
   });
+  const throwAttemptError = (): void => {
+    if (attemptError) throw attemptError;
+  };
   const wrapped = {
     ...base,
     spawn(cmd: string, args: readonly string[], options: ClaudeSpawnOptions) {
@@ -6158,16 +6169,18 @@ function twoAttemptSpawn(
       stdins.push((base.lastChild() as unknown as { stdinChunks?: string[] })?.stdinChunks ?? []);
       return handle;
     },
-    argvs,
-    stdins,
-    assertAttemptHooksSucceeded() {
-      if (attemptHookError) throw attemptHookError;
+    get argvs() {
+      throwAttemptError();
+      return argvs;
+    },
+    get stdins() {
+      throwAttemptError();
+      return stdins;
     },
   };
   return wrapped as FakeSpawn & {
-    argvs: string[][];
-    stdins: string[][];
-    assertAttemptHooksSucceeded: () => void;
+    readonly argvs: string[][];
+    readonly stdins: string[][];
   };
 }
 
@@ -6225,7 +6238,6 @@ test('narrated tool call: opt-in resumes once with the staged prompt still reada
   });
   const { emit, frames } = collect();
   await backend.handle(narratedToolCallTask(), emit, NEVER);
-  fake.assertAttemptHooksSucceeded();
 
   assert.equal(fake.argvs.length, 2, 'expected exactly one corrective retry');
   // Attempt 1 mints the session, attempt 2 continues it — same id either way.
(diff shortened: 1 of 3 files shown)


packages/client/src/backends/claude.test.ts
```diff
@@ -6120,19 +6120,25 @@ function narratedToolCallTask(): TaskAssignFrame {
...
-    setImmediate(() => {
+    const attempt = n++;
+    const lines = scripts[Math.min(attempt, scripts.length - 1)] ?? [];
+    const exitCode = exitCodes[Math.min(attempt, exitCodes.length - 1)] ?? 0;
+    setImmediate(async () => {
+      await beforeAttempt?.(attempt);
       for (const l of lines) child.emitStdout(l.endsWith('\n') ? l : `${l}\n`);
-      setImmediate(() => child.finish(0, null));
+      setImmediate(() => child.finish(exitCode, null));
     });
```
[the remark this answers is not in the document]
longfin: Fixed — the attempt hook now runs through an observed Promise chain. Rejections are caught, the fake child is terminated deterministically, and the test rethrows the captured error via assertAttemptHooksSucceeded so hook failures cannot be silently converted into a passing fallback.

packages/client/src/backends/claude.ts
```diff
@@ -3411,6 +3411,27 @@ export function createClaudeBackend(
           registeredToolNames: [...callerToolNameSet],
         })
       ) {
+        // The first attempt is already a valid completed result. The retry is
+        // speculative: only a newly captured caller tool call is allowed to
+        // replace that outcome. Keep every field that participates in the
+        // terminal decision or response identity so an error/max_turns result
+        // from the corrective child cannot poison the completed turn (#471).
+        const firstAttempt = {
```
longfin: `finalUsage` is missing from the `firstAttempt` snapshot, so the preserved answer is metered with the *discarded* retry's token counts.

Verified on 241ef4a: attempt 1 emits `modelUsage {inputTokens:1000, outputTokens:500}` plus a `completed` result; the retry emits `max_turns`/`is_error` with `modelUsage {inputTokens:7, outputTokens:3}`. The terminal frame comes back as

```
usage = {"promptTokens":7,"completionTokens":3,"model":"claude-fable-5"}
envelope.usage = {prompt_tokens:7, completion_tokens:3, total_tokens:10}
status.message.parts[0].text = "Final report: …"   // attempt 1's answer
```

so attempt 1's 1500 tokens vanish and `usage.model` pairs the *restored* model id with the *discarded* turn's counts. This repo bills x402 per call, so it is a live under-report rather than a cosmetic one.

Fix: add `finalUsage` to both the snapshot and the destructuring restore. (If the retry's spend is meant to be billed too, sum the two attempts explicitly rather than letting the retry silently overwrite.)

packages/client/src/backends/claude.ts
```diff
@@ -3445,6 +3467,33 @@ export function createClaudeBackend(
             `[claude] narrated-tool-call retry failed taskId=${task.taskId} error=${safeToken(errorMessage(err), 300)}`,
           );
         }
+
+        if (capturedToolCalls.length === callerToolCallCountBeforeRetry) {
```
longfin: `emittedAnyArtifact` is not in the restore list, so the `claude-result` fallback artifact that would carry the preserved answer gets suppressed.

Verified on 241ef4a: attempt 1 emits **only** a `result` event (no `assistant` event) with the completed text; the retry emits an assistant text block then `max_turns`. The retry flips `emittedAnyArtifact` to true, the restore leaves it true, so `if (!emittedAnyArtifact && completeText)` (~L3672) never fires. Result:

```
artifacts on the wire = [["claude-message", ["RETRY NOISE"]]]
terminal text         = "Final report: the registered `glob` tool rejected this request."
```

The preserved answer appears *only* on `task.complete`. A consumer that reads the artifact/delta stream and ignores the terminal frame — precisely the case that fallback exists for — receives nothing but the discarded retry's text.

Fix: include `emittedAnyArtifact` in `firstAttempt` and in the destructuring restore.

packages/client/src/backends/claude.ts
```diff
@@ -3445,6 +3467,33 @@ export function createClaudeBackend(
...
+            sawBlockingLimit,
+            assistantModel,
+            initModel,
+            sawCompletedResult,
+            sawErrorResult,
+            sawUnknownToolError,
+            stdoutTail,
+            stderrTail,
+            stdinError,
+          } = firstAttempt);
```
longfin: `streamedResponseText` is not restored, so the streamed artifacts and the terminal envelope end up disagreeing on what the answer was.

Verified on 241ef4a: attempt 1 streams `"Final report: …"`, the retry streams `"RETRY TEXT …"`. `emitAssistantTextUpdate` sees text that does not prefix-match, resets `responseArtifactId`, emits a *second* `claude-message` artifact, and leaves `streamedResponseText` holding the retry's text. After the restore `completeText` is attempt 1's text, so the reconciliation `completeText.startsWith(streamedResponseText)` (~L3683) silently no-ops:

```
artifacts = [["claude-message",["Final report: …"]],
             ["claude-message",["RETRY TEXT that the caller now sees but the envelope drops."]]]
envelope  = content: "Final report: …"
```

A streaming openai-compat caller (reads deltas) and a non-streaming one (reads the envelope) therefore get different answers for the same turn.

Fix: restore `streamedResponseText` (and `responseArtifactId`) with the rest, so the reconciliation branch can emit the corrective tail instead of falling through.

packages/client/src/backends/claude.ts
```diff
@@ -3445,6 +3467,33 @@ export function createClaudeBackend(
...
+            pendingInputRequest,
+            finalText,
+            sawBlockingLimit,
+            assistantModel,
+            initModel,
+            sawCompletedResult,
+            sawErrorResult,
+            sawUnknownToolError,
+            stdoutTail,
+            stderrTail,
```
longfin: Restoring the first attempt's `stderrTail` together with its non-zero `exit` can turn a task that previously **completed** into a hard failure. This one needs an explicit decision more than it needs a code change.

Verified against `pr472^`: attempt 1 writes one benign stderr line (`(node:1) ExperimentalWarning: something`), emits a `completed` result, and exits 1; the retry exits 0 with no tool call.

```
pr472^   → task.complete
241ef4a  → task.fail { code: "claude_exit_nonzero" }
```

because `treatExitAsSuccess` now sees the restored `exit.code === 1` **and** the restored non-empty `stderrTail`.

This matches how the same first attempt behaves with the flag off, so it may well be intended — but it is a user-visible regression for anyone already running `--claude-retry-narrated-tool-call`, and nothing in the diff records the choice. Either keep the retry's exit when it is strictly better than the first attempt's, or state the trade-off in the comment above and pin it with a regression test so it can't be re-litigated by accident.

packages/client/src/backends/claude.test.ts
```diff
@@ -6120,19 +6120,34 @@ function narratedToolCallTask(): TaskAssignFrame {
...
   const argvs: string[][] = [];
   const stdins: string[][] = [];
+  let attemptHookError: Error | null = null;
   let n = 0;
   const base = makeFakeSpawn((child) => {
-    const lines = scripts[Math.min(n, scripts.length - 1)] ?? [];
-    n++;
+    const attempt = n++;
+    const lines = scripts[Math.min(attempt, scripts.length - 1)] ?? [];
+    const exitCode = exitCodes[Math.min(attempt, exitCodes.length - 1)] ?? 0;
```
longfin: `exitCodes` is declared `readonly (number | null)[]`, but `?? 0` coerces an explicit `null` to `0` — so the `null` the type advertises cannot actually be expressed.

Verified on 241ef4a: `twoAttemptSpawn([[NARRATED, OK_RESULT]], [null])` yields `task.complete`, i.e. the fake child reported exit code `0`, not the signal-kill `null` a future test would be writing that for. The backend branches on `exit.code !== 0`, so the two drive materially different paths and the trap is silent.

Fix — keep the repeat-last-element behaviour while letting `null` through:

```ts
const idx = Math.min(attempt, exitCodes.length - 1);
const exitCode = idx >= 0 ? exitCodes[idx] : 0;
```

Or drop `null` from the parameter type if only numbers are ever intended.

packages/client/src/backends/claude.test.ts
```diff
@@ -6145,8 +6160,15 @@ function twoAttemptSpawn(scripts: readonly (readonly string[])[]): FakeSpawn & {
     },
     argvs,
     stdins,
+    assertAttemptHooksSucceeded() {
```
longfin: The hang is genuinely fixed — but `assertAttemptHooksSucceeded()` is opt-in, which leaves the new guard one forgotten call away from a silently green test.

On a hook failure the fake child finishes exit 1 with no stdout; the backend's restore path then still produces `task.complete`, `fake.argvs.length === 2`, and passing `--resume`/argv assertions. Verified on 241ef4a by throwing from the retry hook: **every** other assertion in *narrated tool call: opt-in resumes once…* still holds, and only the explicit `fake.assertAttemptHooksSucceeded()` at L6228 catches it. It is currently called at 1 of the 6 `twoAttemptSpawn` call sites — safe today, since that is the only one passing `beforeAttempt`, but the next author to add a hook inherits the trap.

Fix: make it impossible to forget rather than remembering to call it — e.g. have the wrapped `spawn()` (or the `argvs` / `stdins` accessors) throw when `attemptHookError` is set, so a broken hook fails the test regardless of which assertions the author wrote.

packages/client/src/backends/claude.test.ts
```diff
@@ -6120,19 +6120,34 @@ function narratedToolCallTask(): TaskAssignFrame {
...
     setImmediate(() => {
-      for (const l of lines) child.emitStdout(l.endsWith('\n') ? l : `${l}\n`);
-      setImmediate(() => child.finish(0, null));
+      void Promise.resolve()
+        .then(() => beforeAttempt?.(attempt))
+        .then(() => {
+          for (const l of lines) child.emitStdout(l.endsWith('\n') ? l : `${l}\n`);
+          setImmediate(() => child.finish(exitCode, null));
+        })
+        .catch((err: unknown) => {
```
longfin: Two nits on the new error plumbing.

**The `.catch` covers the script-emission `.then`, not just the hook**, so harness bugs get reported as caller-hook failures. Verified on 241ef4a: feeding a malformed script line makes `assertAttemptHooksSucceeded()` throw `Cannot read properties of null (reading 'endsWith')` — a `twoAttemptSpawn` bug attributed to `beforeAttempt`.

**`attemptHookError` is assigned unconditionally**, so with two failing attempts the *last* error survives rather than the root-cause first one.

Both are cosmetic — the error does surface either way, which is the important part — but the method name misleads during triage. Suggested shape:

```ts
void Promise.resolve()
  .then(() => beforeAttempt?.(attempt))
  .catch((err: unknown) => {
    attemptHookError ??= err instanceof Error ? err : new Error(String(err));
    child.finish(1, null);
    throw err;                 // skip the script emission below
  })
  .then(() => { /* emit lines, finish(exitCode) */ })
  .catch(() => { /* already recorded */ });
```

longfin (commented):
Re-review of `241ef4a`, following up on my review of `8607949`.

**The update correctly fixes the previous `beforeAttempt` finding.** Routing the hook through a promise chain that records the error and calls `child.finish(1, null)` means a throwing hook now lets `backend.handle` resolve promptly instead of stranding the fake child, and `assertAttemptHooksSucceeded()` surfaces the original message. Verified both ways — a hook throwing on attempt 0 and on attempt 1 each fail with the real error (`HOOK BOOM …`) rather than a runner timeout. `tsc --noEmit` is clean and `claude.test.ts` is 186/186 green.

**`packages/client/src/backends/claude.ts` is byte-identical to the previously reviewed head.** `git diff 8607949..241ef4a --stat` touches only `claude.test.ts` (+22/−5). So findings 1–4 below are *unaddressed rather than considered-and-rejected* — flagging that explicitly in case they were simply missed while working through the Copilot pass, not to re-litigate a decision already made.

**Comments 1–3 share a single root cause:** the `firstAttempt` snapshot/restore pair lists a subset of the state `handleEvent` mutates. `finalUsage`, `emittedAnyArtifact`, and `streamedResponseText` are each written by the retry and read after the restore, so each leaks a fragment of the discarded turn into the preserved result — wrong token counts, a suppressed fallback artifact, and streamed-vs-envelope divergence respectively. Worth fixing as one change plus a comment naming the invariant ("every variable `handleEvent` writes that is read after this block must appear in both lists"), which is what would stop the next field from being forgotten.

Comment 4 is a behaviour change I could not tell was intentional; it wants a decision more than a patch. Comments 5–7 are test-harness nits, two of them introduced by this update.

Every scenario above was reproduced by running probe tests against this head, and against `pr472^` where the before/after comparison was the point. Submitting as a comment — none of this blocks the core fix, which does resolve #471 as described.

packages/client/src/backends/claude.ts
longfin: Fixed in b4a2d1b. The retry now isolates finalUsage, then explicitly sums attempt 1 and attempt 2 (including cached/reasoning breakdowns) before emitting protocol usage and the OpenAI-compatible envelope. The new regression test pins 1000+7 prompt and 500+3 completion tokens.

packages/client/src/backends/claude.ts
longfin: Fixed in b4a2d1b. Retry artifacts are now buffered until the retry outcome is accepted. When the first completed result is preserved, the retry buffer is discarded and emittedAnyArtifact is restored, so a result-only first attempt emits the intended claude-result fallback. Added a regression test for this exact case.

packages/client/src/backends/claude.test.ts
longfin: Fixed in b4a2d1b. Hook rejection and script emission now have separate error branches, the state is named attemptError rather than hook-specific, and nullish assignment preserves the first/root-cause error. Script emission exceptions are still surfaced without being misattributed.

packages/client/src/backends/claude.test.ts
longfin: Fixed in b4a2d1b by dropping null from exitCodes; this helper only scripts numeric exits, so its type no longer advertises an outcome that nullish coalescing would silently convert to zero.

packages/client/src/backends/claude.ts
longfin: Fixed in b4a2d1b with the compatibility choice made explicit. We preserve attempt 1 only when it would independently pass the existing terminal-success decision. If attempt 1 is exit 1 plus stderr and the retry exits 0, the retry outcome remains authoritative. A regression test pins that behavior.

packages/client/src/backends/claude.test.ts
longfin: Fixed in b4a2d1b. The opt-in assertion method is gone. argvs and stdins are now accessors that automatically rethrow the recorded attempt error, following the suggested impossible-to-forget shape for helper observations. Added a focused test proving a rejected hook cannot remain silently green.

packages/client/src/backends/claude.ts
longfin: Fixed in b4a2d1b. Speculative retry artifacts no longer reach the wire before outcome selection, and the first-attempt snapshot now includes streamedResponseText, responseArtifactId, and reasoningArtifactId. The regression test asserts the retry text is absent from artifacts while the terminal envelope retains the first answer.

PR #490, opened 2026-09-01
feat: implement Mentionable OAuth federation profile

### LINKED ISSUE #487: identity: implement the Mentionable OAuth federation profile — federated allowed callers, STS token exchange, task authorization
> **This body is the authoritative implementation spec as of 2026-09-01.** It integrates the settled planetarium/mentionable#618 profile (including both v0.1 amendments) and the v0.1 simplification agreed in the comments below. Where a comment below conflicts with this body — in particular the earlier **actor-only task continuity** note, which amendment 2 superseded — **this body wins.**

## Goal

Implement the vicoop-bridge (STS + resource server) half of the Mentionable OAuth federation profile: a Connector exchanges a short-lived, DID-signed assertion for an agent-scoped access token at the bridge's token endpoint (RFC 8693), and every A2A operation — message send and body-less task operations alike — is authorized by that token against server-owned federated allowed callers and task bindings.

The issuer/client half is already implemented and merged: **`@mentionable/connector-kit`** (planetarium/mentionable#620) ships the wire constants, assertion minting, the reference verifiers, and conformance fixtures. Those artifacts are the normative contract — this issue implements against them, not against a prose spec.

## Authoritative decision stack (read all before implementing)

1. planetarium/mentionable#618 — issue body (roles, security model, non-goals).
2. [Decisions 1–4 + proposed defaults](https://github.com/planetarium/mentionable/issues/618#issuecomment-5423799057) — JWT-only assertions, DPoP optional/deferred, no task-scoped down-exchange, scope registry approach.
3. [Decisions 5–9](https://github.com/planetarium/mentionable/issues/618#issuecomment-5449266164) — no refresh tokens (renewal = re-exchange), canonical resource = the agent's advertised A2A endpoint URL, RFC 8414 + Agent Card extension discovery, `private_key_jwt` client auth.
4. [v0.1 amendment 1](https://github.com/planetarium/mentionable/issues/618#issuecomment-5475772893) — **direct Connector delegation only**: `subject_token.iss == authenticated client_id == derived actor`; no `actor_token`; gateway/broker topologies deferred.
5. [v0.1 amendment 2](https://github.com/planetarium/mentionable/issues/618#issuecomment-5476688882) — **task-bound continuation assertions** replace the bare self-assertion for task-scope re-exchange, and the **universal principal-match rule** for task operations. This supersedes amendment 1's actor-only continuity wording and the corresponding comment on this issue.
6. planetarium/mentionable#620 — the merged implementation; its README documents the topology and STS obligations.

## The v0.1 contract (summary — the connector-kit constants are normative)

**Topology** (amendment 1): one directly connected Connector per exchange.

```text
subject assertion issuer == OAuth client_id == derived actor  (all = Connector DID)
relay-verified method    == delegation (fixed; no purpose field, no actor_token)
```

**Message path** — platform-subject assertion (`typ: mentionable-subject-assertion+jwt`; `iss` = Connector DID, `sub` = canonical platform subject such as `slack:T123/U456`, `mentionable_method` claim, `aud` = token endpoint, 300s/600s/60s TTL profile, single-use `jti`) exchanged with `scope` containing message scopes (task scopes may accompany). Derived context: `principal` = platform subject, `actor` = Connector DID.

**Task path** (amendment 2) — task-bound continuation assertion (`typ: mentionable-task-continuation-assertion+jwt`; `sub` = the **originating platform subject**, mandatory `mentionable_task_id` claim, same TTL/`jti` profile, re-minted by the Connector at each renewal). Task scopes ONLY — any message scope with a continuation subject is `invalid_request`. Derived context additionally carries the verified `task_id`. The bare Connector self-assertion is **removed** and must not be accepted.

**Universal task-operation rule** (amendment 2): every task operation — **regardless of whether the presented token carries a task binding** — MUST require `principal == the task's bound principal` in addition to the actor match. A message-path token legitimately carrying task scopes is not exempt. A token whose grant carried a `task_id` is additionally restricted to that task.

**Client authentication**: `private_key_jwt` (`typ: mentionable-client-assertion+jwt`, `client_id` = Connector DID, `iss == sub == client_id`, `aud` = token endpoint, single-use `jti`), verified against the Connector's `did:web` document (`kid ∈ assertionMethod`, EdDSA only).

**Other settled rules**: no refresh tokens (renewal is re-exchange); canonical `resource` = the agent's advertised A2A endpoint URL and is REQUIRED (missing/mismatched → reject); any `actor_token`/`actor_token_type` present → `invalid_request`; scopes restricted to the v0.1 registry (`a2a:message.send`, `a2a:message.stream`, `a2a:task.read`, `a2a:task.cancel`, `a2a:task.resubscribe`, `a2a:task.push-config`), unknown scope → `invalid_scope`; duplicate single-valued params → `invalid_request`; bearer tokens only (DPoP deferred).

## Authorization model: federated allowed callers (no policy object)

Per the v0.1 simplification agreed in the comments (and confirmed by amendment 1 removing the purpose question): **no `identity_policy JSONB`, no purpose field, no grant table.** The one server-owned authorization decision is an exact, collision-safe federated principal tuple registered as an allowed caller of the target agent:

```text
(issuer DID, authentication method URN, canonical platform subject)
e.g. (did:web:slack-connector.example, urn:mentionable:auth:slack-workspace-member:v0.1, slack:T123/U456)
```

- Extend the existing principal grammar (`packages/server/src/auth/principal.ts`) with a `federated` kind. The canonical serialization must be unambiguous — escape or length-prefix components; naive delimiter concatenation is insufficient. The admin API/CLI may present the structured form while `allowed_callers` stores the canonical string.
- **Trust-before-fetch lookup flow** at the token endpoint:
  1. Parse only enough of the *unverified* subject assertion to get candidate `iss`, method, `sub`.
  2. Canonicalize the tuple; exact lookup in the target agent's server-owned `allowed_callers`.
  3. Not present → reject **without** resolving any DID or fetching keys.
  4. Present → full cryptographic verification (client assertion + subject assertion) via the connector-kit contract below.
  5. Issue the resource- and scope-constrained access token; bind normalized principal/actor (and the matched tuple) to created tasks.
- Unverified claims are a lookup key only; they never establish a principal until full verification succeeds.
- **Context stays context**: the message-bound PlatformIdentityCredential path from #469 is untouched, remains non-authorizing, and never enters `allowed_callers`.
- **Invalidation**: task bindings and issued grants record the matched tuple; removing the allowed caller MUST invalidate subsequent exchanges (including continuation renewals for tasks bound to that tuple) consistently across server instances — follow the existing DB-owned `x402_pricing` propagation pattern.

## Consume, don't reimplement

`@mentionable/connector-kit` (signing subpath) ships the verification machinery the STS needs:

- `verifyTokenExchange(params, options)` — THE entry point: shape gate → `verifyClientAssertion` → subject/continuation assertion verification → verified `iss == client_id` binding → returns `DerivedAuthorizationContext { purpose: 'delegation', principal, actor, task_id? }`. It requires `expectedResource` and enforces every wire rule above.
- `verifyClientAssertion`, `verifyOAuthFederationAssertion` — used by the above; available individually.
- Wire constants (typs, claim names, scope registry, error vocabulary) from the main entry.
- **Conformance fixtures**: `fixtures/oauth-federation/v0.1/` (valid exchanges + security-failure cases with pinned reasons). The bridge STS MUST pass them: accept every `valid/` case, reject every `invalid/` case for its stated reason.

The bridge supplies the deployment-owned pieces: the DID resolver (reuse `SafeDidWebResolver` — SSRF hardening is explicitly the STS's responsibility, connector-kit ships no resolver), the replay store (reuse the Postgres `IdentityReplayStore` pattern for `(iss, jti)` single-use, atomic across instances), and the trusted-issuer set (derived from the agent's federated allowed callers, per the trust-before-fetch flow).

## Scope / deliverables

1. **STS**: RFC 8693 `urn:ietf:params:oauth:grant-type:token-exchange` at the existing `POST /oauth/token` (`packages/server/src/auth/device-flow.ts` owns the endpoint today). `private_key_jwt` client auth. Issue agent-scoped access tokens — the existing opaque `vbc_caller_*` pattern is acceptable — whose server-side record carries `{ principal, actor, scopes, resource, task_id?, matched tuple, expiry }`. Revocation via the existing RFC 7009 `/oauth/revoke`.
2. **Discovery**: RFC 8414 metadata at `/.well-known/oauth-authorization-server` (token endpoint, token-exchange grant, accepted subject token types, `private_key_jwt`, `scopes_supported`). Advertise the `https://mentionable.dev/ns/oauth-federation/v0.1` Agent Card extension (params: `authorization_server`, `resource`) on restricted agents' cards (`packages/server/src/agent-card.ts` already synthesizes bridge-owned extensions).
3. **Federated allowed callers**: principal-grammar extension + canonical serialization; admin API + client CLI operations to add/inspect/remove federated callers; RLS-protected mutation paths; changeset for published CLI changes.
4. **Ingress enforcement**: all three A2A paths (`/agents/:id` JSON-RPC, A2A v1 JSON-RPC, A2A v1 HTTP+JSON — including body-less GET) accept the new bearer tokens in `agentAuthMiddleware`; map scopes to operations (`message/send|stream` → `a2a:message.*`, `tasks/get` → `a2a:task.read`, cancel/resubscribe/push-config accordingly); enforce the universal principal-match rule and `task_id` restriction on every task operation identically across transports.
5. **Task binding**: bind `(principal, actor, matched tuple)` at task creation; enforce on later operations; survive access-token rotation (continuation re-exchange).
6. **caller-context-v2**: populate `principal`/`actor`/`attestations` from the verified exchange context (#482/#488 already landed the frame).
7. **Audit**: PII-safe structured events for exchange acceptance/rejection (reuse the `rej_*` correlation-id pattern) — never log raw assertions or token bytes.
8. **Conformance + tests**: drive the STS against the connector-kit fixture suite; unit/e2e coverage for the acceptance criteria below; existing auth-path test suites unchanged.
9. **Docs**: trust-boundary, operator configuration (registering a federated caller end to end with the Slack Connector as the worked example), local/remote testing.

## Explicitly superseded / deferred (do NOT implement)

- ~~`identity_policy JSONB`~~, ~~per-rule `purpose` field~~, ~~grant table~~ → federated allowed-caller tuple (above).
- ~~Actor-less `authentication` purpose~~, ~~independent `actor_token`~~, ~~presenter ≠ issuer / gateway topologies~~, ~~namespace/prefix subject patterns~~ → deferred per amendment 1.
- ~~Promoting the message-bound VC to a principal at message ingress~~ (the original "Ingress ordering" plan) → authorization arrives ONLY via bridge-issued access tokens; the v0.2 VC carrier stays context-only exactly as today.
- ~~Actor-only continuity for task operations~~ → superseded by amendment 2's universal principal-match rule.
- ~~Bearer-less authentication mechanisms for body-less operations~~ → the access token + continuation re-exchange IS the mechanism.
- Refresh tokens, DPoP/mTLS, task-scoped down-exchange → out of scope per #618 decisions.

## Security invariants

- A DID signature alone never grants access; only a fully verified exchange whose exact `(issuer, method, subject)` tuple is an allowed caller of the target agent can.
- Trust policy before network fetch: an unregistered tuple triggers no DID resolution.
- The issuer is authenticated as the OAuth client (`private_key_jwt`); `subject_token.iss == client_id` on verified claims.
- A token issued for one agent/resource cannot authorize another (`resource` bound and required).
- A task ID, context ID, credential ID, or raw assertion is never proof of authority.
- Every task operation requires principal match against the task's bound principal — cross-user access within one Connector fails at the bridge.
- Raw credentials, assertions, and token bytes never reach task storage, WS frames, logs, or backend prompts.
- Replay protection (`(iss, jti)` single-use, both assertion types) is atomic across instances.
- Context-only attestations can never satisfy `allowed_callers`.
- The LLM/backend never performs principal resolution or authorization.

## Acceptance criteria

- [ ] A Connector with an active federated allowed caller obtains a message-scope token via `private_key_jwt` + platform-subject assertion and successfully sends a message on all three ingress paths; the created task binds `(principal = platform subject, actor = Connector DID)`.
- [ ] The same credentials fail for an agent without the tuple; `iss != client_id`, unknown scope, present `actor_token`, expired/replayed assertion, wrong/missing `resource` all fail closed with the profile's error mapping.
- [ ] A continuation-assertion exchange yields a task-scoped token carrying the verified `task_id`; task get/cancel/resubscribe succeed across access-token rotation without any fresh platform event.
- [ ] Per-user isolation: a token derived from another user's continuation assertion — or a message-path token with task scopes for another user — cannot read or cancel the first user's task (universal principal-match).
- [ ] Removing the federated allowed caller invalidates subsequent exchanges and continuation renewals consistently across server instances.
- [ ] The bridge STS passes the `@mentionable/connector-kit` conformance fixture suite (all `valid/` accepted, all `invalid/` rejected for their stated reasons).
- [ ] RFC 8414 metadata and the Agent Card `oauth-federation/v0.1` extension are served and sufficient for a connector-kit client to discover and complete an exchange end to end.
- [ ] `caller-context-v2` delivers principal/actor/attestations for federated exchanges.
- [ ] Context-only PlatformIdentityCredential behavior and every existing authentication path pass their current test suites unchanged.

Related: #467, #469, #482, #488, planetarium/mentionable#618, planetarium/mentionable#619, planetarium/mentionable#620

longfin:
## Summary

- add a profile-neutral RFC 8693 token-exchange core on the existing `/oauth/token` endpoint
- dispatch normalized A2A operations through a resource-authorization profile registry instead of hard-coding Mentionable at ingress
- implement Mentionable OAuth federation v0.1 with DID trust-before-fetch, assertion/client verification, task continuity, and exact caller-tuple policy
- bind principal, actor, profile, and authorization key atomically in the initial task INSERT while keeping internal metadata out of persisted task JSON
- enforce a streaming 64 KiB token-request limit even without trustworthy Content-Length and map DB/profile failures to audited OAuth `server_error` responses
- aggregate RFC 8628 device grant metadata into RFC 8414 discovery when device flow is enabled
- persist and enforce resource, scope, principal, actor, caller tuple, profile, and task bindings across all A2A ingress variants
- add caller-policy propagation, agent-card discovery, admin API/CLI support, schema changes, and operator documentation
- vendor the exact upstream connector-kit verifier and conformance fixtures from commit `f32c8898c7d81b254ec9a562ea2892525db14de6` because the package is not currently published

## Validation

- clean monorepo build: `pnpm -r build`
- monorepo typecheck: `pnpm -r typecheck`
- server tests: 508 total, 433 passed, 75 DB-environment skips, 0 failed
- client tests: 922 passed, 0 failed
- connector-kit exchange fixtures driven through the bridge `/oauth/token` route
- federated message-token integration across v0.3 JSON-RPC, v1 JSON-RPC, and v1 HTTP+JSON, including initial task binding
- synthetic non-Mentionable issuance and resource-authorization dispatch
- chunked oversize, DB failure, and profile failure regression tests
- `git diff --check`
- `pnpm changeset status`

Closes #487


longfin:
## Review @ e69ad64 — LGTM, no blockers; one warning + three nits before merge

Two full review passes (at `578bae3` and again after the token-exchange/profile refactor at `e69ad64`). Verified, with evidence:

- **Vendored connector-kit fidelity**: `oauth-federation-verifier.ts` / `-exchange.ts` / `oauth-federation.ts` and the entire fixture tree are byte-identical to upstream mentionable `f32c8898`; no wire constant is re-declared locally (`profile` re-exports), so contract drift is structurally impossible. The CI-fix and refactor commits never touched `vendor/`.
- **Universal principal-match (amendment 2)**: `task.principalId === caller.principalId` is enforced regardless of whether the token carries a `task_id`, on all three ingress paths (v0 JSON-RPC, A2A v1 JSON-RPC, A2A v1 HTTP+JSON including body-less GET) — the cross-user task access case fails with `task_principal_mismatch`, and there is a direct test for the message-path-token-with-task-scopes case. Task bindings are first-write-wins (`COALESCE`) from server-stamped values only; caller-supplied `_`-prefixed metadata is stripped first.
- **Profile contract**: trust-before-fetch (unregistered tuple ⇒ zero DID resolutions, asserted in tests), `actor == iss == client_id` on verified claims, self-assertion rejected, continuation = task scopes only with `task_id` from the verified payload, `resource` canonicalized and mandatory, `actor_token` presence rejected, duplicate params rejected, unknown scope rejected.
- **Federated tuple encoding**: byte-length-prefixed components — no delimiter injection/collision path found.
- **Issuance/revocation**: token issuance SQL itself guards `allowed_caller = ANY(agents.allowed_callers)`, and verification recomputes `policy_active` per request from the DB — removal of the tuple invalidates exchanges, continuation renewals, and task ops across instances without cache dependence.
- **Replay**: single-transaction `ON CONFLICT DO NOTHING RETURNING` with throw-rollback, consumed only after full verification, digests profile-namespaced.
- **The refactor strengthened things**: `profileId` binding on both tokens and tasks (`unsupported_token_profile` / `task_profile_mismatch`) closes cross-profile confusion pre-emptively; no invariant from the first pass was lost.
- No regressions to the existing `vbc_caller_*` / SIWE / apikey / device-flow paths; context-only PlatformIdentityCredential behavior untouched.

### [WARNING] Conformance fixtures never cross the STS HTTP boundary (AC #6 only transitively met)

`oauth/profiles/mentionable-conformance.test.ts` drives the fixtures against the **vendored library** (`verifyTokenExchange` / `verifyOAuthFederationAssertion`) directly, while `token-exchange/routes.test.ts` exercises the real `POST /oauth/token` handler only with hand-minted assertions (zero fixture references). The STS's own pre-verifier plumbing — candidate decode, `agentIdFromResource`, `allowed_callers` dispatch, `typ` routing, and (post-refactor) the core→profile hop — is therefore never exercised by the `invalid/` fixtures with their pinned reasons. Issue #487's acceptance criterion reads "the bridge STS passes the fixture suite"; as written that is only satisfied transitively. Not exploitable (the endpoint delegates to the byte-identical verifier and the critical paths have independent tests), but recommend adding a test that POSTs each `exchange-request-*` fixture's params to the mounted endpoint and asserts HTTP status + `error` per the manifest, before merge.

### [NITS]

1. `oauth/token-exchange/types.ts` and the profile's `subjectTokenTypes` re-declare three RFC URN strings locally instead of importing the vendored constants. Values verified identical today; importing removes even theoretical drift.
2. The `TokenExchangeProfile` contract allows a future profile to return `replays: []` and silently opt out of replay protection. A core-side guard (reject `ok` results with empty `replays` unless the profile explicitly declares statelessness) would make the invariant structural rather than reviewed-in.
3. The `oauth_federation_*` → `oauth_token_exchange_*` schema rename ships no migration/`DROP` for the interim tables — harmless if `578bae3` was never deployed anywhere; worth a note otherwise.

Issue #497, opened 2026-09-09
[Umbrella] Implement caller-scoped execution runtimes with Docker


longfin:
## Objective

Track delivery of caller-scoped execution environments using the existing Docker runtime, while preserving one public agent identity and one bridge connection.

Design background: #496. Provider comparison and Docker-first recommendation: https://github.com/planetarium/vicoop-bridge/issues/496#issuecomment-5594436695.

This is the implementation umbrella for the Docker path. #496 remains the broader design/requirements reference. Implementation has not started under this tracker; the stages below are planned work, not completed validation. Add child issue and PR links to each stage as work is split out.

## Scope and rollout

- Preserve existing host/single-runtime behavior as an explicit supported mode. Caller isolation is opt-in.
- First vertical slice: Docker + Claude plain A2A + directly authenticated principals.
- Missing identity, unsupported negotiation, and delegated requests without an implemented authorization policy fail closed in isolated mode.
- Extend the same routing/provider structure to Codex with a separate backend/app-server per execution scope.
- Add execution-bound MCP/artifact support after the core runtime boundary works. Reject unsupported feature combinations explicitly in intermediate releases.
- OpenSandbox implementation, OpenClaw Gateway isolation, and isolation of externally executed caller functions are outside this Docker tracker. Preserve the existing contract that caller-provided function invocations are returned upstream for execution.

## Execution model

```text
Server authentication and authorization
  -> transport-owned execution scope
  -> CallerScopedBackend
      -> acquire scope's Docker runtime + backend + execution lease
      -> stage input files and translate paths
      -> backend.handle(task, emit, signal)
      -> confirm execution cleanup and release lease
```

- **Execution scope:** agent identity plus authorized caller and, when supported, delegation/grant boundary. Never selected by prompts, tool arguments, or caller-provided runtime names.
- **Conversation scope:** execution scope plus contextId.
- **Execution generation:** taskId/executionId, retained through cancellation and replay.
- Use stable, opaque, collision-resistant names within Docker/runtime naming limits. Validate ownership labels and mounts before adopting existing resources.

## Decisions to close before the first integrated release

- [ ] Define the server-owned scope wire contract, capability negotiation, policy versioning, and unsupported-client behavior.
- [ ] Define direct-principal behavior for public agents and shared credentials; shared credentials identify a shared principal, not individual humans.
[...]

PR #499, opened 2026-09-10
feat(client): add R2 caller-isolated Claude Docker runtime

longfin:
Implements the restricted R2 release of #497 on top of R1 (#498). Opting into `--backend claude --runtime caller-container` gives directly authenticated callers separate workspaces and live conversation bindings while retaining one agent identity and bridge connection. This PR targets `codex/497-docker-runtimes`; it does not merge or deploy to `main`.

The server authenticates callers even for an empty allowlist, advertises the required authentication, derives the scope, and rejects anonymous/delegated execution before payment/allocation. The client requires explicit R2 negotiation and validates the agent/principal/scope/generation before allocation. Existing host and single-container modes remain available.

### Runtime and persistence contract

- Serialize each scope through allocation, execution and confirmed cleanup; bound scope count, queueing, contexts, memory/PIDs, task time and storage. Canceling a waiter does not release another execution's lease. Queued context-capacity rejection preserves existing conversations. Uncertain cleanup quarantines the scope.
- Run each execution in a unique container/private Docker bridge network, with read-only rootfs, bounded tmpfs, no host mounts, non-root jobs and a root watchdog. Freeze and verify workload threads before snapshotting; remove the whole container/network before publishing success.
- **R2 implementation choice:** use separate workspace/HOME/CLI-state directories on bounded tmpfs plus opaque host-owned snapshots, instead of unbounded Docker named volumes. Successful checkpoints persist across executions/restarts. Failure/cancellation discards uncommitted changes; cancellation after atomic commit retains the checkpoint and reports that distinction. Response loss is not exactly-once execution.
- Recover dead owners and remove stale resources before accepting work. Restart restores files and explicitly reports a fresh conversation. `caller-state` provides exclusive offline inspection/deletion. Detached `stop` honors the recorded launch-time cleanup budget (120 seconds plus exit margin) and retains the pidfile until cleanup ends.
- Require an immutable backend-installed image and explicit credential provisioning: the existing private API-key file (default), or `credentialSource: "host-claude"` with no `credentialFile`. Host OAuth mode reads the default macOS Keychain/Linux credential file per execution and forwards only the access token. No refresh token, operator settings or conversations are copied. Invalid/expired/near-expiry credentials fail closed; refresh is operator-managed. Workloads can access the supplied credential.

Supported: Linux/macOS Docker + Claude plain A2A, text and inline image/PDF inputs. Delegation, isolated Codex/OpenClaw, OpenAI compatibility/caller tools, MCP, URI fetching and outgoing file delivery remain unavailable. Includes image recipe, operator configuration/limits, server-first rollout/rollback guidance and a client-only minor changeset. Details: [`docs/caller-runtime.md`](docs/caller-runtime.md).

### Validation

- Repository build/typecheck passed; final client build/typecheck passed after config validation changes.
- Client: **959 passed**. Server: **454 passed / 78 skipped**. Protocol: **11 passed**.
- Real Docker under tsx and a Bun-compiled smoke: A/B/A files, snapshot restoration/rollback, explicit env, bounded storage, cancellation/descendant cleanup and separate caller networks.
- Actual pinned Node 22 + Claude **2.1.265** image built and used for the Bun runtime smoke. Image-default workdir permissions were caught and fixed during this validation.
- Full Bun-compiled CLI with a **deterministic Claude fixture**: strict argv/prompt-file staging, A/B/A live sessions/workspaces, WebSocket disconnect/replay, SIGKILL after an observed uncommitted write, restart/reset and orderly shutdown. The actual `start --detach` / `stop` path also passed with Docker removal delayed 12 seconds, a post-launch config change, and no forced kill or leftover resources. This full-CLI fixture does not call the model.
- Real host OAuth through `DockerCallerRuntimePool`: Claude 2.1.265 / Haiku inference, checkpoint and container/network cleanup passed under tsx and a Bun-compiled smoke. This validates the runtime authentication path, not full A2A OAuth E2E or expiration/refresh. Full compiled-client fixture regression also passed after the change.
- Adversarial security/lifecycle subagent review and follow-up fixes: queue admission accounting, commit/cancel reporting, reasoning option propagation and public-agent authentication advertisement. Follow-up adversarial review of `cc0733e` verified the context-limit/session-preservation and detached-stop grace fixes, with no additional confirmed defects in those changes.
- One initial parallel test run stalled in the unchanged OpenClaw suite; its standalone 80 tests and subsequent complete client runs passed.

Not deployed or released. R3 conversation restoration/idle eviction and R4–R6 features remain separate work.

OAuth follow-up has local validation above; the earlier adversarial review predates this follow-up. No new adversarial review is claimed.

### DIFF
--- a/.changeset/tidy-caller-isolation.md
+++ b/.changeset/tidy-caller-isolation.md
@@ -3,3 +3,5 @@
 ---
 
 Add opt-in Claude caller-isolated Docker execution with bounded transactional workspaces, direct-principal negotiation, cancellation cleanup, crash reconciliation, and offline state inspection/deletion. Requires an upgraded bridge server and a pinned backend image; existing host and single-container modes remain supported.
+
+Preserve existing conversations when queued requests exceed the context limit, and let detached caller daemons finish their cleanup before `stop` escalates to SIGKILL.
--- a/.changeset/tidy-caller-isolation.md
+++ b/.changeset/tidy-caller-isolation.md
@@ -5,3 +5,5 @@
 Add opt-in Claude caller-isolated Docker execution with bounded transactional workspaces, direct-principal negotiation, cancellation cleanup, crash reconciliation, and offline state inspection/deletion. Requires an upgraded bridge server and a pinned backend image; existing host and single-container modes remain supported.
 
 Preserve existing conversations when queued requests exceed the context limit, and let detached caller daemons finish their cleanup before `stop` escalates to SIGKILL.
+
+Allow caller-isolated Claude to reuse the operator's host OAuth login through explicit `host-claude` credential provisioning, forwarding only the current access token per execution. Existing API-key-file configuration remains supported; token refresh remains operator-managed.
--- a/.changeset/tidy-caller-isolation.md
+++ b/.changeset/tidy-caller-isolation.md
@@ -1,0 +1,5 @@
+---
+"@vicoop-bridge/client": minor
+---
+
+Add opt-in Claude caller-isolated Docker execution with bounded transactional workspaces, direct-principal negotiation, cancellation cleanup, crash reconciliation, and offline state inspection/deletion. Requires an upgraded bridge server and a pinned backend image; existing host and single-container modes remain supported.
--- a/docs/caller-runtime-rollout.md
+++ b/docs/caller-runtime-rollout.md
@@ -3,6 +3,9 @@
 Tracking: [#497](https://github.com/planetarium/vicoop-bridge/issues/497).
 Design background: [#496](https://github.com/planetarium/vicoop-bridge/issues/496).
 
+R2 implementation and operating contract: [Claude caller runtime](caller-runtime.md).
+The R1 section below describes the earlier release boundary.
+
 ## R1: compatible foundations
 
 R1 supports the existing `host` and `container` modes. Their install,
--- a/docs/caller-runtime.md
+++ b/docs/caller-runtime.md
@@ -160,7 +160,9 @@ A crash during the short owner-file critical section may leave `.guard`; inspect
 owner PID, Docker resources and directory ownership before manually removing this
 lock directory. A wedged Docker engine can leave a quarantine/owner record and
 resources; restore Docker access and restart for reconciliation. Orderly shutdown
-awaits tasks/cleanup for up to 120 seconds, then process exit leaves recovery to the
+awaits tasks/cleanup for up to 120 seconds. Detached `stop` honors the launch-time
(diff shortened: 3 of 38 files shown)


longfin:
Requirement clarification: R2 should **allocate a dedicated Docker container per authenticated user and reuse it across that user's requests**.

```text
One public agent / bridge connection
├── Alice authenticated → Alice's dedicated container
│                          ├── dedicated workspace
│                          └── dedicated Claude/Codex conversation state
└── Bob authenticated   → Bob's dedicated container
                           ├── dedicated workspace
                           └── dedicated Claude/Codex conversation state
```

- Allocation is keyed by the server-verified `agent identity + authenticated principal`. Neither the model nor request arguments may select the user or container.
- Subsequent requests from the same user, including requests with a new `contextId`, route to that user's existing dedicated container. Conversations are distinguished by `contextId` within that user's scope.
- Different users do not share containers, workspaces, or backend conversation state.
- Preserve workspace and conversation state in dedicated per-user volumes. After stopping/restarting or recreating a container, attach only that user's storage. Explicitly handle whether conversation resumption is supported and what happens if it fails.
- Processes and temporary authentication grants may still have per-execution lifetimes. The host authentication broker can be shared, while real provider credentials remain outside user containers. Reuse the shared execution/authentication infrastructure from #501 and #505 where applicable.

**The current PR's approach—create a temporary container for each execution, restore a user snapshot, checkpoint on success, then remove the container—does not match this expectation.** Separating only the persisted state by user is not sufficient to satisfy this requirement. Please revise R2 around the creation, reuse, stopping, and recovery of dedicated per-user containers.

Preserve the existing safety requirements: confirmed execution termination on cancellation, protection of active work, deduplicated concurrent allocation, resource limits, and restart recovery. In particular, cancellation, cleanup, or recovery for one user must not affect another user's container.

Validation should demonstrate that Alice and Bob receive distinct containers; Alice's subsequent requests and new conversations reuse Alice's existing container; and stopping/recreating containers preserves access exclusively to each user's own files and conversation state.

This comment clarifies the isolation model. It does not expand R2's Claude-first scope to include Codex, or require all later operational features such as automatic idle reclamation to ship in R2.

longfin:
Superseded by #506 following the dedicated per-user container requirement clarified above. The replacement draft starts from current main (including #501 and #505) and records the revised design and implementation/acceptance checklist. It currently contains documentation only; the replacement runtime and fresh validation remain outstanding. Applicable scope/authentication and lease/queue work from this PR can be reused selectively. This PR is closed without merging; its execution-container/snapshot design and validation remain historical reference. #497 remains open.

